PR #5, opened 2021-05-02
Replace memmap with memmap2

upsuper:
The crate [memmap](https://crates.io/crates/memmap) has been marked unmaintained, see the [advisory](https://rustsec.org/advisories/RUSTSEC-2020-0077).

This PR replaces the dependency with [memmap2](https://crates.io/crates/memmap2), which is a maintained fork of memmap.


msirringhaus:
Thank you!

PR #9, opened 2022-01-19
Add the number of threads that we were tried to suspend to the corresponding error

gabrielesvelto:
…


msirringhaus:
Looking good! Thanks!

PR #10, opened 2022-02-08
port to nix 0.23

NickCao:


msirringhaus:
LGTM
Is there a related bugzilla-entry for the version bump yet?

NickCao:
Not aware of bugzilla entries, just me trying to fix build of firefox 97 on riscv.

msirringhaus:
Yes, but merging here is only the first step. This crate has to be updated [in Firefox](https://searchfox.org/mozilla-central/source/third_party/rust/minidump_writer_linux) as well, which needs a bugzilla-entry. (Unless you don't want to upstream your fixes and only build with local patches)

PR #14, opened 2022-03-14
Refactor layout

Jake-Shadle:
Part 1 of splitting #13.

This refactors the layout of the crate and cleans up imports a bit, but no functional changes.

- Puts the linux/android specific code into a `linux` submodule that is reexported from the root so the outward API doesn't really change, it will just make adding additional target platforms cleaner
- Renamed `LinuxPtraceDumper` to `PtraceDumper`, the naming was inherited from breakpad but doesn't make much sense since there is only one ptrace implementation, and even if a future one was added eg BSD, it would either be its own separate one, or else a specialization of the linux one
- There are quite a few target specific parts of this code requiring relatively complicated cfg expressions, so I added `cfg-if` to make them cleaner and easier to read
- A lot of the imports where massive with lots of duplication, so I collapsed most of them
- Cleaned up the names of nested modules as they were needlessly prefixed with the parent module name
- Removed all of the empty mips/ppc/sparc implementations

One breaking change I made was to move to the 2021 edition which is 5 months old at this point, but I can revert that if it's not wanted, the only change that this really enabled was to remove all of the `std::convert::TryInto` imports since that is part of the prelude in the 2021 edition.


Jake-Shadle:
Yah I'm not a fan of it either particularly, it was just the lazy option for sections since they shared so many imports between the different submodules.

PR #16, opened 2022-03-14
Cleanup

Jake-Shadle:
Follow up to #15, this bumps outdated dependencies (no changes needed), and fixes all clippy lints, at least when targeting x86_64

PR #17, opened 2022-03-15
Use unified crash-context

Jake-Shadle:
This builds on #16 to use the `CrashContext` from [here](https://github.com/EmbarkStudios/crash-handling/blob/impl/crash-context/src/linux.rs) as it allows a single crash context to be shared between different crates, without needing to rely on `minidump-writer` itself, and implements getcontext for targets that don't directly support it such as musl and android.

Resolves: #4

### DIFF
--- a/src/linux/dumper_cpu_info/arm.rs
+++ b/src/linux/dumper_cpu_info/arm.rs
@@ -26,7 +26,7 @@ pub fn parse_cpus_from_sysfile(file: &mut File) -> Result<HashSet<u32>> {
             continue;
         }
         let cores: std::result::Result<Vec<_>, _> =
-            items.split("-").map(|x| x.parse::<u32>()).collect();
+            items.split('-').map(|x| x.parse::<u32>()).collect();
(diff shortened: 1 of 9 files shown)


src/linux/dumper_cpu_info/arm.rs
```diff
@@ -235,51 +255,36 @@ pub fn write_cpu_information(sys_info: &mut MDRawSystemInfo) -> Result<()> {
...
 
-            elf_hwcaps
-        };
+    // The sys_info.cpu field is just a byte array, but in arm's case it is
+    // actually
+    // minidump_common::format::ARMCpuInfo {
+    //  pub cpuid: u32,
+    //  pub elf_hwcaps: u32,
+    // }
+    use scroll::Pwrite;
```
msirringhaus, line 272: Sorry for the nitpicking again :-D

msirringhaus (commented):
Sorry, one more nit.

msirringhaus (approved):
Thanks!

PR #19, opened 2022-03-29
Add initial Windows implementation

Jake-Shadle:
- Add first pass at windows minidump writer
- Remove assertion info, it is functionally useless
- Always open process handle internally
- Use published crash-context

See https://github.com/EmbarkStudios/crash-handling/pull/9 where it is being used.


Jake-Shadle:
Oops, looks like minidump has changed, I'll fix this tomorrow.

Jake-Shadle:
Oh, and add windows tests as well.

Jake-Shadle:
Hmm, so I go the in process dump test working easily, but the external one doesn't work due to some problems internally in MiniDumpWriteDump failing to do `ReadProcessMemory` for some reason. My assumption is that the child process is maybe exiting early, or at least some variable is falling out of scope somehow, but couldn't figure it out. I'm not super concerned with that as my own tests only use external process dumping and they have worked flawlessly https://github.com/EmbarkStudios/crash-handling/tree/main/minidumper-test

marti4d:
Hi @Jake-Shadle. Aria asked me to help do a first-pass review of this, as she is currently very busy on another project. I should some review comments ready for you soon. Sorry about the wait!

Jake-Shadle:
No worries, thanks!

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl MinidumpWriter {
+    /// Creates a minidump writer for a crash that occurred in an external process.
+    pub fn external_process(
+        crash_context: crash_context::CrashContext,
+        pid: u32,
+    ) -> Result<Self, Error> {
+        let crashing_process = unsafe {
+            OpenProcess(
+                GENERIC_ALL, // desired access
```
marti4d, line 43: Is there a reason for preferring this instead of `PROCESS_ALL_ACCESS`?

It's interesting that this value works, as I don't see it mentioned in the [MSDN docs for OpenProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess). `PROCESS_ALL_ACCESS` also has the advantage of being in `Win32_System_Threading`, so you wouldn't have to define the special constant above

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Creates a minidump writer for a crash that occurred in the current process.
+    ///
+    /// # Errors
+    ///
+    /// Fails if we are unable to open a `HANDLE` to the current process
+    pub fn current_process(crash_context: crash_context::CrashContext) -> Result<Self, Error> {
+        let crashing_pid = std::process::id();
+
+        // SAFETY: syscall
+        let crashing_process = unsafe {
```
marti4d, line 70: You could use [`GetCurrentProcess`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocess) instead. It never fails and always has `PROCESS_ALL_ACCESS`

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+                ExceptionPointers: self.crash_context.exception_pointers as *mut _,
+                ClientPointers: if self.is_external_process { 1 } else { 0 },
+            })
+        } else {
+            None
+        };
+
+        // This is a bit dangerous if doing in-process dumping, but that's not
+        // (currently) a real target of this crate, so this allocation is fine
+        let mut user_streams = Vec::with_capacity(2);
```
marti4d, line 107: If this will interfere with in-process dumping, does in make sense to remove the `current_process()` function for now?

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+                validity: BreakpadInfoValid::DumpThreadId.bits()
+                    | BreakpadInfoValid::RequestingThreadId.bits(),
+                dump_thread_id: self.crash_context.thread_id,
+                // Safety: syscall
+                requesting_thread_id: unsafe { GetCurrentThreadId() },
+            };
+
+            // TODO: derive Pwrite for MINIDUMP_BREAKPAD_INFO
+            // https://github.com/rust-minidump/rust-minidump/pull/534
+            let mut offset = 0;
```
marti4d, line 129: nit - You could use gwrite for these and it will update the offset for you

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+
+            user_streams.push(md::MINIDUMP_USER_STREAM {
+                Type: MINIDUMP_STREAM_TYPE::BreakpadInfoStream as u32,
+                BufferSize: breakpad_info.len() as u32,
+                // Again with the mut pointer
+                Buffer: breakpad_info.as_mut_ptr().cast(),
+            });
+        }
+
+        let handle_stream = self.fill_handle_stream();
```
marti4d, line 142: I have a suggestion that you can feel free to ignore: I think it would make the code a bit more clear to just have `fill_handle_stream()` return an `Option<Vec<u8>>` and move the creation of the `md::MINIDUMP_USER_STREAM` to the "if let" statement below, like so:

```
if let Some(buf) = &mut handle_stream {
    user_streams.push(md::MINIDUMP_USER_STREAM {
        Type: MINIDUMP_STREAM_TYPE::HandleOperationListStream as u32,
        BufferSize: buf.len() as u32,
        // Still not getting over the mut pointers here
        Buffer: buf.as_mut_ptr().cast(),
    });
}
```


That way it keeps the subfunction focused on creating the stream buffer and this function continues to focus on building up the user stream headers. It also more closely mirrors what you already did above with the `breakpad_info`

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        // Note that we do this by ref, as the buffer inside the option needs
+        // to stay alive for as long as we're writing the minidump since
+        // the user stream has a pointer to it
+        if let Some((_buf, handle_stream)) = &handle_stream {
+            user_streams.push(*handle_stream);
+        }
+
+        let user_stream = md::MINIDUMP_USER_STREAM_INFORMATION {
```
marti4d, line 151: Perhaps `user_stream_info` would be a bit more clear of a name? Seems a bit strange that `user_stream` contains `user_streams` and not the other way around

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+        } else {
+            Ok(())
+        }
+    }
+
+    /// In the case of a `STATUS_INVALID_HANDLE` exception, this function
+    /// enumerates all of the handle operations that occurred within the crashing
+    /// process and fills out a minidump user stream with the ops pertaining to
+    /// the last invalid handle that is enumerated, which is, presumably
+    /// (hopfully?), the one that led to the exception
```
marti4d, line 184: typo 'hopfully'

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+        // SAFETY: syscall
+        if unsafe {
+            av::VerifierEnumerateResource(
+                self.crashing_process,                // process to enumerate the handles for
+                0,                                    // flags
+                av::AvrfResourceHandleTrace,          // resource typea, we want to trace handles
+                Some(enum_callback),                  // enumeration callback
+                (&mut hs as *mut HandleState).cast(), // enumeration context
+            )
+        } != 0
```
marti4d, line 230: Isn't '0' a success code? Maybe I'm misunderstanding somehow, but this seems like it would run the "good path" when the call is unsuccessful?

src/windows/minidump_writer.rs
```diff
@@ -0,0 +1,279 @@
...
+                SizeOfEntry: std::mem::size_of::<av::AVRF_HANDLE_OPERATION>() as u32,
+                NumberOfEntries: 0,
+                Reserved: 0,
+            };
+
+            stream_buf.resize(md_list.SizeOfHeader as usize, 0);
+
+            #[inline]
+            fn to_bytes<T: Sized>(v: &T) -> &[u8] {
+                let s = std::slice::from_ref(v);
```
marti4d, line 246: I don't think the first part is necessary - A single expression of `std::slice::from_raw_parts((v as *const T) as *const u8, std::mem::size_of::<T>())` seems a bit more idiomatic?

tests/windows_minidump_writer.rs
```diff
@@ -0,0 +1,140 @@
...
+    let mut tmpfile = tempfile::Builder::new()
+        .prefix("windows_current_process")
+        .tempfile()
+        .unwrap();
+
+    unsafe {
+        let mut exception_record: EXCEPTION_RECORD = mem::zeroed();
+        let mut exception_context = mem::MaybeUninit::uninit();
+
+        RtlCaptureContext(exception_context.as_mut_ptr());
```
marti4d, line 41: It might be worth asserting that this succeeds, since if it doesn't it will invoke UB in the next line

marti4d (commented):
This is very nice, and it's great that you also wrote tests for it! Thanks a bunch for this, and sorry about the long wait for this review.

Most of my comments below are just minor suggestions and nitpicks -- You can probably ignore them if you don't agree; however, I have 3 concerns:

1) I think there may be reversed logic in `windows/minidump_writer.rs#230`
2) You mention that `let mut user_streams = Vec::with_capacity(2);` may interfere with in-process dumping. If that's the case, does it make sense that we are providing a `current_process()` function to allow a behavior that won't work properly?
3) I saw your comments above, and for the record I also see the failure of the `dump_external_process` test with the same code. Does it make sense to disable that test until it's fixed?

WRT to the failing test, I've done a bit of investigation and it looks like if you pass `std::ptr::null()` in place of `exc_info` during the call to `MiniDumpWriteDump`, it works... So the issue seems to be stemming from that.

It's getting a bit late for me right now, but tomorrow I will continue investigating --It looks like it *should* work, so it's fairly intriguing that it doesn't.

src/windows/minidump_writer.rs
Jake-Shadle: This was just the access that the original Breakpad code was using and it just worked so actually didn't even look at the documentation for this function, but can try changing it.

src/windows/minidump_writer.rs
Jake-Shadle: This was just an offhand comment that might seem more ominous than it is, I don't think there's any reason to not have current process dumping, I'll just add a comment on it that users should note in-process dumping is less reliable just in general.

src/windows/minidump_writer.rs
Jake-Shadle: Oops.

tests/windows_minidump_writer.rs
Jake-Shadle: `RtlCaptureContext` doesn't have a return value unfortunately.

src/windows/minidump_writer.rs
Jake-Shadle: Yup seems to work just fine.

src/bin/test.rs
```diff
@@ -1,292 +1,357 @@
...
+        // are
+        unsafe {
+            let mut exception_record: EXCEPTION_RECORD = mem::zeroed();
+            let mut exception_context: CONTEXT = mem::zeroed();
+            exception_context.ContextFlags =
+                minidump_common::format::ContextFlagsAmd64::CONTEXT_AMD64_ALL.bits();
+
+            let tid = GetCurrentThreadId();
+
+            let thread_handle = OpenThread(THREAD_ALL_ACCESS, 0, tid);
```
marti4d, line 322: nit: There also exists a [`GetCurrentThread()`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentthread) Win32 API that never fails and returns a handle that doesn't need to be closed and has `THREAD_ALL_ACCESS` permissions

marti4d (commented):
Thanks for addressing the feedback so quickly! I have a few more optional suggestions, but as far as I'm concerned this code is ready to check in once we figure out why that test is broken.

I'm investigating that right now, so hopefully we can get that resolved quickly and merge this change 😀

marti4d:
@Jake-Shadle I left a PR to fix the broken test on your Embark Studios fork. Once you merge that in, I think this PR should be good-to-go! 😀

src/bin/test.rs
```diff
@@ -1,292 +1,358 @@
...
+            let mut exception_context: CONTEXT = mem::zeroed();
+            exception_context.ContextFlags =
+                minidump_common::format::ContextFlagsAmd64::CONTEXT_AMD64_ALL.bits();
+
+            let pid = GetCurrentProcessId();
+            let tid = GetCurrentThreadId();
+
+            let thread_handle = GetCurrentThread();
+            GetThreadContext(thread_handle, &mut exception_context);
+            CloseHandle(thread_handle);
```
marti4d, line 325: nit: You don't have to use `CloseHandle` on this because `GetCurrentThread()` returns a pseudohandle and MSDN guarantees that `CloseHandle` is a no-op.

So you could replace all 3 lines with this one-liner:

`GetThreadContext(GetCurrentThread(), &mut exception_context);`

marti4d (approved):
@Gankra Hey Aria! IMO this change is good-to-go and I'm confident to merge it in at anytime.

Thanks a lot, @Jake-Shadle. This is great :)

PR #21, opened 2022-04-20
Add initial MacOS implementation

Jake-Shadle:
This essentially copies the Breakpad minidump creation, with a couple of minor changes.

1. Breakpad uses `getrusage` to determine the process start time, however...this was wrong. `getrusage` can only retrieve the process information for the current task (process) or its children, this implementation instead uses `proc_pidinfo` to get the process info for the actual crashed process.
2. Breakpad/Crashpad both retrieve OS/kernel version information via an deprecated and undocumented (that's not saying much though, Apple hates documentation it seems) method, however there is a perfectly fine way to retrieve this information via `sysctlbyname`, `kern.osproductversion` has been available since `10.13.4` which is 2 versions older than the oldest supported version, and `kern.osversion` has been available for...I actually don't know how long, again, Apple is really terrible at documentation.

Note that this implementation only supports MacOS (not iOS), and only `x86_64` and `aarch64` as those are the only officially supported architectures on MacOS now.

Also note, most of my local testing has been done on an aarch64 machine, I'll add building of aarch64 from an x86_64 to CI to cover that case, but we won't be able to run actual tests (not that CI actually runs tests atm...) until Github Actions adds M1 runners, which who knows when that will happen.

- https://github.com/actions/virtual-environments/issues/2187

Another option would be for Embark to supply a temporary M1 runner for this project similarly to how we already supply one for https://github.com/bytecodealliance/wasmtime as seen in https://github.com/EmbarkStudios/wasmtime-aarch64-apple-darwin.


Jake-Shadle:
Converting this to a draft, I've just found out that I completely misunderstood `mach_task_self` which always returns the same value, I had assumed it was a unique identifier like a PID, but that's completely wrong and apparently I need to refactor the code I was using to test this change (in addition to the tests in this PR).

Jake-Shadle:
I've now got this working on x86_64 and aarch64, I've commented out the test that I used for confirmation since its dependency chain leads to openssl and the build was failing and I don't want to deal with it at the moment, I'll fix it properly tomorrow by opening a PR to `breakpad-symbols` to use rustls instead of openssl.

Gankra:
Is this something that the vendored openssl hack fixes? https://github.com/rust-minidump/rust-minidump/blob/main/minidump-stackwalk/Cargo.toml#L27

Jake-Shadle:
Possibly, but the dependency was on minidump-processor which doesn't have that feature. Exposing the rustls feature on reqwest or making reqwest/http symbols optional altogether would definitely work though.

Gankra:
yeah in theory all the crates should have the hack but I just tossed on the one I needed to ship

src/mac/mach.rs
```diff
@@ -0,0 +1,571 @@
...
+    kern_return::{kern_return_t, KERN_SUCCESS},
+    port::mach_port_name_t,
+    task::{self, task_threads},
+    task_info,
+    thread_act::thread_get_state,
+    traps::mach_task_self,
+    vm::{mach_vm_deallocate, mach_vm_read, mach_vm_region_recurse},
+    vm_region::vm_region_submap_info_64,
+};
+
```
Jake-Shadle, line 15: The mach2 crate has a really annoying layout that I supposed they inherited from Apple.

src/mac/mach.rs
```diff
@@ -0,0 +1,571 @@
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let sz = self.count as usize;
+        (sz, Some(sz))
+    }
+}
+
+/// Retrieves an integer sysctl by name. Returns the default value if retrieval
+/// fails.
+pub fn sysctl_by_name<T: Sized + Default>(name: &[u8]) -> T {
+    let mut out = T::default();
```
Jake-Shadle, line 493: `sysctl` integers can be 32 or 64 bits, this just lets you retrieve either.

src/mac/minidump_writer.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::io::{Seek, Write};
+
+type Result<T> = std::result::Result<T, WriterError>;
+
+pub struct MinidumpWriter {
+    /// The crash context as captured by an exception handler
+    pub(crate) crash_context: crash_context::CrashContext,
+    /// List of raw blocks of memory we've written into the stream. These are
+    /// referenced by other streams (eg thread list)
+    pub(crate) memory_blocks: Vec<MDMemoryDescriptor>,
```
Jake-Shadle, line 16: This is the only reason that `dump` is `&mut`, but we could do the same thing that the Linux dumper does and just create an internal mut state instead if that signature is problematic.

src/mac/streams/system_info.rs
```diff
@@ -0,0 +1,200 @@
...
+fn os_version() -> (u32, u32, u32) {
+    let vers = mach::sysctl_string(b"kern.osproductversion\0");
+
+    let inner = || {
+        let mut it = vers.split('.');
+
+        let major: u32 = it.next()?.parse().ok()?;
+        let minor: u32 = it.next()?.parse().ok()?;
+        let patch: u32 = it.next().and_then(|p| p.parse().ok()).unwrap_or_default();
+
```
Jake-Shadle, line 20: The version is _usually_ a standard semver `<major>.<minor>.<patch>` but it _can_ be just `<major>.<minor>`

src/mac/streams/thread_list.rs
```diff
@@ -0,0 +1,231 @@
...
+                // specifies 32 bits for the flags register, we can truncate safely
+                // with no loss.
+                out.eflags = ts.__rflags as _;
+                out.cs = ts.__cs as u16;
+                out.fs = ts.__fs as u16;
+                out.gs = ts.__gs as u16;
+            } else if #[cfg(target_arch = "aarch64")] {
+                // This is kind of a lie as we don't actually include the full float state..?
+                out.context_flags = format::ContextFlagsArm64Old::CONTEXT_ARM64_OLD_FULL.bits() as u64;
+
```
Jake-Shadle, line 219: The MacOS threadstate doesn't include floating point registers on aarch64, but Breakpad was still setting `MD_CONTEXT_ARM64_FULL_OLD` which includes `MD_CONTEXT_ARM64_FLOATING_POINT_OLD`. From what I have seen this is fine and stackwalker/sentry have no issues with the thread state in such a dump.

src/mem_writer.rs
```diff
@@ -0,0 +1,272 @@
+use crate::minidump_format::{MDLocationDescriptor, MDRVA};
+use scroll::ctx::{SizeWith, TryIntoCtx};
+
```
Jake-Shadle, line 3: The contents of this file were moved to not be Linux specific.

src/windows/minidump_writer.rs
```diff
@@ -13,8 +13,6 @@ pub struct MinidumpWriter {
     crash_context: crash_context::CrashContext,
     /// Handle to the crashing process, which could be ourselves
     crashing_process: HANDLE,
-    /// The pid of the crashing process.
-    crashing_pid: u32,
     /// The `EXCEPTION_POINTERS` contained in crash context is a pointer into the
```
Jake-Shadle, line 16: This was a slight refactoring from the bump to `crash-context 0.2` since the pid is now part of the crash-context itself on Windows to match linux and mac, so this allowed us to just cleanup the interface a bit since it can internally open the process in the correct way without the user needing to specify it.

tests/mac_minidump_writer.rs
```diff
@@ -0,0 +1,194 @@
...
+
+//     assert!(
+//         fake_crash_thread.frames.iter().any(|sf| {
+//             sf.function_name
+//                 .as_ref()
+//                 .map_or(false, |fname| fname.ends_with("wait_until_killed"))
+//         }),
+//         "unable to locate expected function"
+//     );
+// }
```
Jake-Shadle, line 194: This is temporarily disabled, but I think this would be a good way to test all the platforms, making the test binary more complicated, and verify that minidump-process finds the expected threads and stackframes. As noted, the dump_syms issue where it is only packaged as a binary right now makes it annoying to use in CI at the moment, and breakpad symbols depends on openssl which unnecessarily complicates builds, especially since this is all local symbol resolution. (at least for now)

Jake-Shadle:
FYI I've filed https://github.com/mdsteele/rust-cab/pull/13 to get rid of the 2 security advisories introduced by this change.

Issue #23, opened 2022-04-28
MacOS follow up


Jake-Shadle:
#21 adds MacOS support but while testing I noticed a few things that _might_ be improved, though it's hard to say given the embarrassing state of Apple's documentation. Writing this down now so I don't forget later.

- Minidump supports thread names, however the MacOS implementation doesn't write that data, because Breakpad doesn't. This _might_ (again, terrible documentation) be impossible since the thread name is actually part of pthreads, but on the minidump-writer side we are working with mach ports which are (mostly?) paired with a pthread, but AFAICT the API to retrieve the thread name for the pthread is task specific and probably won't work? But this would be extremely nice information to have if we can get it reliably. There might be some shenanigans to be done by finding out where the BSD pthread info is located in the process space and just reading it directly from the process memory like we do with eg the thread stack etc.

gabrielesvelto:
We've already implemented this in Breakpad, see: https://hg.mozilla.org/mozilla-central/rev/63e43c6f4114

Porting it should be straightforward.

PR #32, opened 2022-05-25
Add MacOS thread names

Jake-Shadle:
I wanted to use `proc_pidinfo` for this since apparently it is possible, but was apparently missing something and the calls failed (maybe need to convert the mach port to a BSD/posix thread id?), so for now this uses the mach `thread_info` to retrieve the thread names.

Note this depends on a change to dump_syms that I noticed when testing, will file a PR for that now as well.

Resolves: #23

### DIFF
--- a/src/mac/streams/thread_names.rs
+++ b/src/mac/streams/thread_names.rs
@@ -73,7 +73,7 @@ impl MinidumpWriter {
             thread_info.pth_name.as_ptr().cast(),
             thread_info.pth_name.len(),
         )))
-        .map_err(|err| TaskDumpError::from(err))?;
+        .unwrap_or_default();
 
         // Ignore the null terminator
         let tname = match name.find('\0') {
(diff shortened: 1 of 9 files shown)


src/mac/streams/thread_names.rs
```diff
@@ -0,0 +1,91 @@
...
+        dumper: &TaskDumper,
+    ) -> Result<MDRawDirectory, WriterError> {
+        let threads = dumper.read_threads()?;
+
+        // Ignore the thread that handled the exception
+        let thread_count = if self.crash_context.handler_thread != mach2::port::MACH_PORT_NULL {
+            threads.len() - 1
+        } else {
+            threads.len()
+        };
```
gabrielesvelto, line 18: It'd be nice to put this in a helper function with a name that describes the intent so we can share it with `MinidumpWriter::write_thread_list()` and drop the comment.

gabrielesvelto (changes requested):
Looking good, requesting just a tiny change

gabrielesvelto (approved):
LGTM!

PR #38, opened 2022-06-30
Fix stack pointer recording for aarch64 Linux

sfackler:
Register 13 is the stack pointer on 32 bit ARM, but on 64 bit ARM it's stored separately.


Jake-Shadle:
Thanks!

Jake-Shadle (approved):

Issue #30, opened 2022-05-25
Run tests in CI


Jake-Shadle:
Right now we only test that the code builds (for a wide variety of targets) and passes lints, but we don't actually run any tests which is...not great. We should at least run tests for the major targets, with the exception of `aarch64-apple-darwin` since github doesn't have M1 runners, still.

PR #48, opened 2022-07-18
Run tests in CI

Jake-Shadle:
There's a lot of warning spam due to unused code etc, so cleaning those up so CI is nicer.

Resolves: #30


Jake-Shadle:
Had to disable some tests in CI, unsure why the linux ones fail since they run fine locally, the mac ones probably fail simply because the CPU allocated for macos runners is abysmal so the tests time out.

Issue #39, opened 2022-07-06
The Linux minidump writer is mishandling EINTR when waiting for a thread to stop after having attached to it


gabrielesvelto:
This is something I noticed reading the code, see the [loop here](https://github.com/rust-minidump/minidump-writer/blob/7bf892e4ab0a7109d3bcc88746897e179e0e74eb/src/linux/ptrace_dumper.rs#L116-L121). We've just called `ptrace(PTRACE_ATTACH, pid)` on a given thread and we need to `waitpid()` for it. If we get an error and `errno` is `EINTR` (that is we received a signal and handled it) we're supposed to wait again, if we get any other error we should detach from the thread and return. It seems to me that we're doing the opposite (i.e. we detach and return on `EINTR` and try again for other values of `errno`).

PR #49, opened 2022-07-19
Retry attaching to a process with ptrace::attach() if the call is interrupted, bail out otherwise

gabrielesvelto:
Resolves: #39


Jake-Shadle (approved):

msirringhaus:
lgtm

Jake-Shadle:
It feels like we should setup Github auto merge or mergify or something, especially now that we are actually running tests in CI.

PR #53, opened 2022-08-17
Make `dir_section` and `mem_writer` modules public

sage-msft:
I plan to construct minidumps from other data sources, so having access to types provided by these modules would be helpful.

Based on the current public API it seems the primary use case of `minidump-writer` is to construct a minidump by introspecting on the current execution environment.

Could you support this use case, either by adopting this change or working with me to find another solution?

### DIFF
--- a/examples/synthetic.rs
+++ b/examples/synthetic.rs
@@ -1,0 +1,36 @@
+//! Emits default minidump with no streams to specified path
+
+use std::fs::File;
+
+use minidump_writer::{
+    dir_section::DirSection,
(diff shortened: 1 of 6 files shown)


gabrielesvelto:
This is fine by me, @Jake-Shadle @Gankra have you any opinions on this?

Jake-Shadle (approved):
I guess this is fine for now, but might make more sense to pull out the helpers around writing the minidump to the minidump crate at some point since it feels a bit weird to compile a bunch of code meant for inspecting external process state without using any of it.

PR #59, opened 2022-10-24
Use the winapi crate instead of the windows-sys one for the Windows bindings

### LINKED ISSUE #58: Remove `windows-sys` dependency
`windows-sys` is still moving far too quickly resulting in churn and duplicate dependencies, the usage in this crate is quite limited, so it makes sense to just bite the bullet and handwrite the binding we actually need and remove the dependency so that downstream crates don't have to care which version of windows-sys they may use.

gabrielesvelto:
This fixes #58

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,18 +31,9 @@ nix = { version = "0.25", default-features = false, features = [
     "user",
 ] }
 
-[target.'cfg(target_os = "windows")'.dependencies.windows-sys]
-version = "0.42"
-features = [
-    # MiniDumpWriteDump requires...a lot of features
-    "Win32_Foundation",
-    "Win32_Storage_FileSystem",
-    "Win32_System_Diagnostics_Debug",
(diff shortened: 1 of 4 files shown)


gabrielesvelto:
@Jake-Shadle this falls back to the older winapi crate plus a bunch of bindings which aren't in that crate and which I've rolled directly inside minidump-writer. We could also add all bindings directly inside minidump-writer and drop winapi too given the bindings are small. Let me know WDYT

Jake-Shadle:
My plan was to just put the bindings in the crate, using winapi feels like taking a step back in some way, especially since one reason it is being supplanted is because at least windows-sys covers the entire api surface, but the cost in churn is not worth it imo.

gabrielesvelto:
Alright, I was thinking myself of just rolling the few bindings right into the crate so that it has no dependencies and has an overall smaller footprint. Will do that later today if I have some spare cycles.

PR #60, opened 2022-10-26
Remove the windows-sys dependency

gabrielesvelto:
Resolves: #58

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,5 @@
 fn main() {
-    #![cfg(all(target_os = "windows"))]
-    println!("cargo:rustc-link-lib=dylib=dbghelp");
+    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("windows") {
+        println!("cargo:rustc-link-lib=dylib=dbghelp");
+    }
 }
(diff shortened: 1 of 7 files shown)


gabrielesvelto:
The bindings are lifted from windows-sys, save for the structures used in minidump callbacks which are replaced with dummies (we don't use them and they use a zillion different types). This isn't great because I have to make the bindings public in order to use them from the tests; I'm not sure if there' s a way to expose them just to the test but not the the crate's users.

build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    #![cfg(all(target_os = "windows"))]
+    println!("cargo:rustc-link-lib=dylib=dbghelp");
+}
```
Jake-Shadle, line 4:
```suggestion
    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("windows") {
        println!("cargo:rustc-link-lib=dylib=dbghelp");
    }
}
```

Jake-Shadle (changes requested):
Just one minor change needed to ensure cross compilation works. Not sure why the build is freaking out about build.rs though

Jake-Shadle:
Oh nvm, it's because it's an outer attribute so applies to the main function so my change would also fix the build issue.

Jake-Shadle (approved):
:fire:

PR #66, opened 2022-12-06
Update the rust-minidump dependencies

gabrielesvelto:
This makes the minidump-writer crate not depend anymore on the tracing crate (it becomes only a dev dependency) which reduces the footprint of the crate when it's being vendored.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ byteorder = "1.3.2"
 cfg-if = "1.0"
 crash-context = "0.5"
 memoffset = "0.7"
-minidump-common = "0.14"
+minidump-common = "0.15"
 scroll = "0.11"
 tempfile = "3.1.0"
 thiserror = "1.0.21"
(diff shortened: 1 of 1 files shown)


Jake-Shadle (approved):

Jake-Shadle:
@gabrielesvelto I don't have admin on this repo, it would be nice if you could setup mergify or something so that we don't have to manually merge PRs once they've been reviewed, I always forget on this repo since every other repo I maintain has that setup so I don't have to think about it.

gabrielesvelto:
I'm not familiar with it but I'll try looking how you did it in your repos, I'm always happy to reduce the number of manual chores.

Jake-Shadle:
Here's the typical one we use https://github.com/EmbarkStudios/cargo-deny/blob/main/.mergify.yml

sylvestre:
@JohanLorenzo is an expert in this

gabrielesvelto:
@JohanLorenzo I will, this can be useful here, in the [rust-minidump](https://github.com/rust-minidump/rust-minidump) repo as well as in the [dump_syms](https://github.com/mozilla/dump_syms/) one.

PR #68, opened 2023-01-05
Don't fail when ptrace::detach returns ESRCH.

### LINKED ISSUE #29: Don't fail when `ptrace::detach()` returns an `ESRCH` error
If one of the threads we've suspended disappears while we're writing the minidump the call to `ptrace::detach()` in `PtraceDumper::resume_thread()` will fail setting `errno` to `ESRCH`. This is a completely benign error which we can ignore since we simply don't care about the state of the process/thread after we're done with it. The same applies to pretty much any other use of `pthread::detach()`, we could ignore this error for all of them.

afranchuk:
ESRCH indicates that the thread has disappeared while we had it stopped. We don't consider this an error.

Closes #29.

### DIFF
--- a/src/linux/ptrace_dumper.rs
+++ b/src/linux/ptrace_dumper.rs
@@ -50,6 +50,22 @@ impl Drop for PtraceDumper {
     }
 }
 
+/// PTRACE_DETACH the given pid.
+///
+/// This handles special errno cases (ESRCH) which we won't consider errors.
+fn ptrace_detach(child: Pid) -> Result<(), DumperError> {
+    let pid = nix::unistd::Pid::from_raw(child);
+    ptrace::detach(pid, None).or_else(|e| {
+        // errno is set to ESRCH if the pid no longer exists, but we don't want to error in that
+        // case.
+        if e == nix::Error::ESRCH {
+            Ok(())
+        } else {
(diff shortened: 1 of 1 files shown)


afranchuk:
Note that we could probably add a test to cover this case; let me know if that's desirable (the existing tests fail on my local machine sometimes, I am going to look into that).

gabrielesvelto:
If you have time to write a test that's always welcome! Also it seems that recent versions of clippy are complaining about [redundant casts](https://github.com/rust-minidump/minidump-writer/actions/runs/3849683432/jobs/6583915787), could you address it in a separate commit within this PR? Thanks for your contribution!

gabrielesvelto (approved):
Actually never mind the test, it would be highly timing dependent and we already have a way of detecting if this work in our usage in Firefox. This LGTM.

gabrielesvelto:
I fixed the clippy issues after merging your PR, I didn't want to waste your time on them given they were trivial. As for the test I thought about it and it's a lot of work for little benefit: it's a scenario we routinely hit in production and we have a dedicated path for handling errors. If it doesn't work as expected we'll see it in production, if it works then the error will be gone and we get more minidumps. So there's no downsides to not having test coverage.

PR #70, opened 2023-02-24
Write a MemoryInfoListStream on linux.

### LINKED ISSUE #8: Evaluate writing out a MemoryInfoListStream
To improve stack walking heuristics we'd like to validate ambiguous instruction pointers by checking if they fall within an executable reason. On Windows this is already possible because **windbg.dll** populates the `MemoryInfoListStream` with the whole state of the process' memory when it snapshots it. On Linux we currently have `/proc/self/maps` which contains the required information but we'd have to parse it in the stackwalker to yield the same information. Alternatively we might try and populate the `MemoryInfoListStream` on Linux too in order to be able to use the same code across different platforms.

Populating the stream should be relatively simple, info regarding the [header](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_memory_info_list) and [entries](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_memory_info) are readily available. The biggset issue is that the `AllocationProtect`, `State`, `Protect` and `Type` fields are supposed to contain Windows-specific constants. There's two ways to tackle this: either we use those fields to store Linux-specific constants - but that would defeat the purpose of having identical code for all platforms in the stack walker - or we map them to their closest Windows equivalents. Interestingly crashpad does the latter for [fuchsia](https://chromium.googlesource.com/crashpad/crashpad/+/d8f60c81124fd2a9f3e64e59c0ae2a18e6700a12/snapshot/fuchsia/memory_map_region_snapshot_fuchsia.cc#26) but it does not populate the stream on Linux and macOS.

afranchuk:
This makes it easier for consumers to read the minidumps, rather than needing to parse linux-specific information.

Closes #8.


afranchuk:
This relies on changes to `minidump-common`, so is a draft for now.

afranchuk:
Since a new release of `minidump-common` was released, this can move out of draft status.

Jake-Shadle:
Thanks, I'll take a look at this tomorrow!

Jake-Shadle (approved):
Oof, this is a big change, but the added/improved testing definitely helps!

afranchuk:
@Jake-Shadle Yeah, I'm not sure if you saw but the changes to the tests exposed a bug in the tests which was hiding a failure (https://github.com/rust-minidump/minidump-writer/pull/70/commits/7ef6250b1f4e7af9df185d62351fda75f903a6e0). For now, I had no solution to the failure.

src/linux/ptrace_dumper.rs
```diff
@@ -374,7 +363,7 @@ impl PtraceDumper {
         // the bitfield length is 2^test_bits long.
         let test_bits = 11;
         // byte length of the corresponding array.
-        let array_size = 1 << (test_bits - 3);
+        let array_size: u64 = 1 << (test_bits - 3);
```
gabrielesvelto, line 366: I think this should be an `usize`, it would remove several redundant casts and ensure we don't have an array_size that's larger than the pointer size on 32-bits hosts.

gabrielesvelto (changes requested):
This is fine save for one small comment I left. The refactoring resulting from using the `procfs` crate is really nice and I also like how you macro-ized the tests.

src/linux/ptrace_dumper.rs
afranchuk: Sounds good to me. To be honest, I'm not sure why I had changed it to an explicit type, previously I believe it was inferred as `usize` based on the usage as an index.

afranchuk:
CI failure seems to be from network errors.

gabrielesvelto (approved):

gabrielesvelto:
I'm re-triggering the jobs, I'll merge the changes as soon as they all turn green

gabrielesvelto:
> @Jake-Shadle Yeah, I'm not sure if you saw but the changes to the tests exposed a bug in the tests which was hiding a failure ([7ef6250](https://github.com/rust-minidump/minidump-writer/commit/7ef6250b1f4e7af9df185d62351fda75f903a6e0)). For now, I had no solution to the failure.

I'm reworking that stuff in https://github.com/rust-minidump/minidump-writer/pull/78. Right now situations in which the stack pointer falls out of a mapped range just don't work.

PR #78, opened 2023-04-28
Write stacks correctly during stack overflows

### LINKED ISSUE #24: Detect guard pages and dump the real stack instead when encountering stack overflows on Linux
After landing bug [1678152](https://bugzilla.mozilla.org/show_bug.cgi?id=1678152) in Firefox we noticed that several of the crash reports we got had an empty stack. The first frame was fine but the rest of the stack appeared to be an empty 4KiB chunk. As it turns out in those crashes the stack pointer had been bumped straight into the guard page below the stack causing [this code](https://searchfox.org/mozilla-central/rev/88792eff309001778cb2431f2a0ed92f8f3c258a/toolkit/crashreporter/breakpad-client/linux/minidump_writer/linux_dumper.cc#791-795) code to mistake the guard page for the stack and dumping it. The new writer [inherited this flaw from Breakpad](https://github.com/rust-minidump/minidump-writer/blob/39a1c8009dd59c4c43d381f6b2a24fa6ae048ef6/src/linux/ptrace_dumper.rs#L320-L325). To fix this we should have `MappingInfo` store the readable/writable flags in addition to the executable one and if the mapping [we find here](https://github.com/rust-minidump/minidump-writer/blob/39a1c8009dd59c4c43d381f6b2a24fa6ae048ef6/src/linux/ptrace_dumper.rs#L330-L332) is non readable/writable then we should shift the stack pointer to the previous page and pick the previous mapping instead.

gabrielesvelto:
When encountering a stack overflow we often crash accessing the guard page. The logic assumed that wherever the stack pointer was so was the stack, but this lead the writer to dump the guard page in these cases. This patch changes the logic to inspect the properties of the mapping that appears to correspond to the stack and - if it looks like a guard page - look for the actual stack instead.

This fixes #24

### DIFF
--- a/src/linux/maps_reader.rs
+++ b/src/linux/maps_reader.rs
@@ -36,8 +36,8 @@ pub struct MappingInfo {
     // address range. The following structure holds the original mapping
     // address range as reported by the operating system.
     pub system_mapping_info: SystemMappingInfo,
-    pub offset: usize,    // offset into the backed file.
-    pub executable: bool, // true if the mapping has the execute bit set.
+    pub offset: usize,              // offset into the backed file.
+    pub permissions: MMPermissions, // read, write and execute permissions.
     pub name: Option<OsString>,
     // pub elf_obj: Option<elf::Elf>,
 }
@@ -78,8 +78,6 @@ impl MappingInfo {
             let end_address: usize = mm.address.1.try_into()?;
             let mut offset: usize = mm.offset.try_into()?;
 
-            let executable = mm.perms.contains(MMPermissions::EXECUTE);
-
             let mut pathname: Option<OsString> = match mm.pathname {
                 MMapPath::Path(p) => Some(p.into()),
                 MMapPath::Heap => Some("[heap]".into()),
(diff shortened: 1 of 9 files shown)


gabrielesvelto:
This is a WIP, I need to test it in real world scenarios and add tests too.

gabrielesvelto:
This is as good as it gets, I tried to match the original intent, removing the redundant limit on the amount of stack we captured while ensuring we always grab a stack, provided the overflow is within 1MiB of where it should be as that's the size of guard pages in Linux ATM.

gabrielesvelto:
@Jake-Shadle would you have time to take a look?

Jake-Shadle:
Oh sorry, this completely fell off my radar, will review tomorrow!

gabrielesvelto:
Thanks!

src/linux/maps_reader.rs
```diff
@@ -339,6 +337,18 @@ impl MappingInfo {
...
+    }
+
+    pub fn is_readable(&self) -> bool {
+        self.permissions.contains(MMPermissions::READ)
+    }
+
+    pub fn is_writable(&self) -> bool {
+        self.permissions.contains(MMPermissions::WRITE)
+    }
 }
```
Jake-Shadle, line 352: These could be inline, but that's just a nit.

Jake-Shadle (approved):
LGTM, can cut a release on Monday unless there is a rush.

gabrielesvelto:
> LGTM, can cut a release on Monday unless there is a rush.

No rush at all.

PR #80, opened 2023-05-17
Use procfs-core rather than procfs to reduce dependencies.

afranchuk:


afranchuk:
@gabrielesvelto this should make vendoring easier.

gabrielesvelto:
FYI I just tested vendoring this and it's fine. We only have to bump a couple more crates but they're either written by us or I can do  the audit myself quickly.

Jake-Shadle (approved):

gabrielesvelto:
Thanks Jake, merging

PR #81, opened 2023-06-07
Stabilize ARM and AArch64 support on both Linux and Android, add support for x86-64 on Android

gabrielesvelto:


Jake-Shadle (approved):
This change looks reasonable, but we really should find a nice way of adding tests, or at least sanity checks of some of this stuff to make reviews easier.

Jake-Shadle:
Well tests are harder for Android, when saying sanity checks I was more thinking just dumping out byte buffers and ensuring they can be interpreted correctly across changes in the code/libc/nix, but obviously has much lower value. But might be worth at least mentioning that Firefox is running more tests in addition to those in the repo.

PR #83, opened 2023-06-13
Increase the ThreadState structure minimum alignment to 8 bytes

### LINKED ISSUE #82: The `ThreadState` structure only has 4-bytes alignment which causes issues when cast into types with stricter alignment requirements
As per title this is causing issues, see [this](https://github.com/EmbarkStudios/crash-handling/actions/runs/5202171455/jobs/9383396374?pr=76#step:7:120).

gabrielesvelto:
This fixes #82

### DIFF
--- a/src/mac/mach.rs
+++ b/src/mac/mach.rs
@@ -234,6 +234,7 @@ cfg_if::cfg_if! {
     }
 }
 
+#[repr(C, align(8))]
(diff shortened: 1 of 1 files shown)


Jake-Shadle (approved):
Nice!

gabrielesvelto:
Thanks

PR #84, opened 2023-06-19
Update

Jake-Shadle:
Updates all dependencies, and reverts #80 until `procfs-core` is released. Requested in https://github.com/eminence/procfs/issues/273.

Also fixed the `test_merged_mappings` which was failing on my machine due to the ` (deleted)` suffix causing the path comparison to fail. It now just checks the beginning matches which should effectively be the same unless this crate's testing gets far more comprehensive in the future.


Cargo.toml
```diff
@@ -32,7 +32,9 @@ nix = { version = "0.26", default-features = false, features = [
     "user",
 ] }
 # Used for parsing procfs info.
-procfs-core = { git = "https://github.com/eminence/procfs", rev = "591016286d195f523628eccf5cac9c923ae6ea45" }
+procfs = "0.15.1"
+# TODO: move to when it is published as it reduces dependencies
+#procfs-core = { git = "https://github.com/eminence/procfs", rev = "591016286d195f523628eccf5cac9c923ae6ea45" }
```
gabrielesvelto, line 37: I'd rather have us revert https://github.com/rust-minidump/minidump-writer/commit/b95bf2e31f0be429a78f9639835312d8423fd632 instead of falling back on vanilla `procfs` because we can't vendor it in Firefox due to its broad dependencies (including windows-sys which we can't still vendor). If that's not a problem I'll re-land the switch to `procfs-core` as soon as they cut a proper release.

Cargo.toml
```diff
@@ -13,16 +13,16 @@ bitflags = "2.0"
...
 scroll = "0.11"
 tempfile = "3.1.0"
 thiserror = "1.0.21"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.74"
-goblin = "0.6"
-memmap2 = "0.5"
+goblin = "0.7"
+memmap2 = "0.7"
```
gabrielesvelto, line 25: rust-minidump seems to be pulling in 0.5.7 and that's part of the reason for the failures on macOS, I'm going to update it

Cargo.toml
```diff
@@ -44,13 +46,13 @@ mach2 = "0.4"
...
-memmap2 = "0.5"
+minidump = "0.17.0"
+memmap2 = "0.7"
 
 [target.'cfg(target_os = "macos")'.dev-dependencies]
 # We dump symbols for the `test` executable so that we can validate that minidumps
 # created by this crate can be processed by minidump-processor
 dump_syms = { version = "2.0.0", default-features = false }
-minidump-processor = { version = "0.16.0", default-features = false }
+minidump-processor = { version = "0.17.0", default-features = false }
```
gabrielesvelto, line 56: We need to add `minidump-unwind` as a dependency as some of the functionality we're using in the tests has been split out into that crate.

gabrielesvelto (changes requested):

Cargo.toml
Jake-Shadle: Ahh ok I'll wait for a response from the maintainers (probably US? timezone) today and if they can't make a release can revert that.

Cargo.toml
gabrielesvelto: Or we do the opposite and go back to 0.5.x until they update first.

Jake-Shadle: Yah that's what I opted for, easiest.

Cargo.toml
```diff
@@ -13,15 +13,15 @@ bitflags = "2.0"
 byteorder = "1.3.2"
 cfg-if = "1.0"
 crash-context = "0.6"
```
gabrielesvelto, line 15: You just need to bump crash-context to `0.6.1` and we're done, thank you!

gabrielesvelto (changes requested):

Cargo.toml
Jake-Shadle: 0.6.1 is semver compatible with 0.6, I tend to leave off compatible versions since it just leads to churn in the manifest file, but I'll add it for clarity.

gabrielesvelto (approved):

eminence:
Hi all, just dropping a quick note to say that `procfs-core` v0.16.0 has now been published.  You are currently depending on v0.16.0-RC1.   The difference between the two is very small, just documentation fixes, and two new feature (parsing `/proc/net/snmp` data and `/proc/partitions` data) that won't impact you

PR #85, opened 2023-06-26
Avoid pulling in `chrono`

Jake-Shadle:
By default `procfs-core` pulls in `chrono` which we've actually banned since it had vulnerabilities and was not actively maintained (though now is), but is regardless completely unneeded for what we use `procfs-core` for in this crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,8 @@ nix = { version = "0.26", default-features = false, features = [
     "user",
 ] }
 # Used for parsing procfs info.
-procfs-core = "0.16.0-RC1"
+# default-features is disabled since it pulls in chrono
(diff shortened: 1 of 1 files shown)


gabrielesvelto:
Thanks, this will help us too!

repi:
possible to do a new release of the crate to include this?

gabrielesvelto:
Yeah, I'm working on this crate this week and I'll cut a new release

gabrielesvelto:
@Jake-Shadle would you have time to cut a new release? I don't seem to be able to. There's also https://github.com/rust-minidump/minidump-writer/pull/87 to consider before cutting a new release (it's fine by me BTW).

PR #87, opened 2023-09-18
update dependencies

demoray:


Jake-Shadle (approved):
Thanks!

demoray:
@Jake-Shadle It would be extremely helpful to release a new version of minidump-writer.

Applications that use this and need to use a newer version of minidump or minidump-common have to jump through hoops to have multiple versions of the same crate.

Thanks for the work on this crate!

Jake-Shadle:
https://github.com/rust-minidump/minidump-writer/releases/tag/0.8.2

PR #89, opened 2023-09-19
Merge empty mappings within a library so they form a single contiguous range

### LINKED ISSUE #88: Merge empty mappings within libraries so that they form a single module
On Android we frequently find that the linker loads Firefox' libxul.so with holes in it. These corresponds to areas that have been reserved by the linker but weren't actually mapped for some reason. See below for an example of how these mappings look like:
```
9b4a0000-9b931000 r--p 00000000 08:12 393449     /data/app/org.mozilla.firefox-1/lib/x86/libxul.so
9b931000-9bcae000 ---p 00000000 00:00 0
9bcae000-a116b000 r-xp 00490000 08:12 393449     /data/app/org.mozilla.firefox-1/lib/x86/libxul.so
a116b000-a4562000 r--p 0594d000 08:12 393449     /data/app/org.mozilla.firefox-1/lib/x86/libxul.so
a4562000-a4563000 ---p 00000000 00:00 0
a4563000-a4840000 r--p 08d44000 08:12 393449     /data/app/org.mozilla.firefox-1/lib/x86/libxul.so
a4840000-a4873000 rw-p 09021000 08:12 393449     /data/app/org.mozilla.firefox-1/lib/x86/libxul.so
```
We already have logic to fold this type of empty pages but it only works if the previous mappings were executable. In this case they aren't, so the writer will emit two modules for libxul.so. One for the `9b4a0000-9bcae000` range and one for the `9bcae000-a4873000` range. These should be merged together instead.

gabrielesvelto:
This issue was observed in Firefox, see bug 1837471. This fixes #88

### DIFF
--- a/src/linux/maps_reader.rs
+++ b/src/linux/maps_reader.rs
@@ -70,6 +70,14 @@ impl MappingInfo {
         is_mapping_a_path(self.name.as_deref())
     }
 
+    pub fn is_empty_page(&self) -> bool {
+        (self.offset == 0) && (self.permissions == MMPermissions::PRIVATE) && self.name.is_none()
+    }
+
+    pub fn end_address(&self) -> usize {
+        self.start_address + self.size
+    }
+
     pub fn aggregate(memory_maps: MemoryMaps, linux_gate_loc: AuxvType) -> Result<Vec<Self>> {
         let mut infos = Vec::<Self>::new();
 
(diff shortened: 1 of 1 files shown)


Jake-Shadle (approved):
Thanks for the test, makes it much easier to review. :)

gabrielesvelto:
Thanks Jake!

Issue #31, opened 2022-05-25
Add soft errors to minidump


Jake-Shadle:
We've discussed loosening many of the hard failures inherited from Breakpad into more soft errors, as in many cases, even if a piece of process information cannot be retrieved for some reason, the dump will still contain all of the useful information that we did retrieve. For example, if we can't retrieve the name of a thread for some reason, or it isn't utf-8, that should just mean we skip that thread or write an empty or placeholder name for it, but don't fail the entire dump creation.

This however means that when you get a minidump it might be missing information you were expecting. What would be cool is if in these soft errors cases actually recorded the failure, eg "failed to get thread name for thread <id>", and actually add those failures to a separate minidump stream at the end. This would allow end users to still get the useful information that was retrieved during minidump creation, but also see exactly why there might be various things missing that they might have expected.

gabrielesvelto:
Adding a datapoint here: most of the times we've seen Firefox fail to generate a minidump it's been for these three reasons (in order of volume):
* Inability to suspend the threads in the target process (issue #28)
* Failing to get the auxv vector (issue #27)
* Failing to access something in `/proc` (with `/proc/cpuinfo` being to be the most affected)

PR #94, opened 2023-10-26
Add support for the `MINIDUMP_HANDLE_DATA_STREAM` to Linux

gabrielesvelto:


gabrielesvelto:
This relies on my changes in PR https://github.com/rust-minidump/rust-minidump/pull/885 so I need to land that one first and cut a new release of the `rust-minidump` crate.

Jake-Shadle:
At lunch now, will take a look when I get back.

src/linux/minidump_writer.rs
```diff
@@ -327,6 +327,9 @@ impl MinidumpWriter {
         // Write section to file
         dir_section.write_to_file(buffer, Some(dirent))?;
 
+        let dirent = handle_data_stream::write(self, buffer)?;
+        dir_section.write_to_file(buffer, Some(dirent))?;
+
```
Jake-Shadle, line 332: I do have a bit of a concern about the potential fallibility of this operation, if it fails we generate no dump at all, even when all previous sections have been dumped successfully, but I don't know if this will matter in real world usage or not. I suppose if Firefox deploys this change and doesn't see it negatively impacting successful dump creation then it's not a big deal, just something I want to bring up.

Actually, I think the linux dumper just needs to be changed so that every section can fallback to being not written if it fails internally due to whatever reason with the hopes that at least some sections can be written so there is at least _something_ being dumped rather than 1 failure leaving the crash with no diagnosable artifact, but that can be discussed separately.

Jake-Shadle (approved):

src/linux/minidump_writer.rs
gabrielesvelto: > Actually, I think the linux dumper just needs to be changed so that every section can fallback to being not written if it fails internally due to whatever reason with the hopes that at least some sections can be written so there is at least _something_ being dumped rather than 1 failure leaving the crash with no diagnosable artifact, but that can be discussed separately.

I agree. I've been monitoring all the possible failure modes when writing minidumps in Firefox and not being able to write a section doesn't happen very often... but it does happen. I can make this particular operation fallible in the meantime.

BTW if we discuss making all the non-necessary sections fallible it would be nice to introduce a mechanism to gather multiple errors. Right now we know what will cause a failure but it's an all-or-nothing situation. It would be nice to have a mechanism so that we get a list of errors to distinguish between scenarios like "we could write the minidump but we don't have the file descriptors" or "we could only partially write a minidump, threads X, Y & Z are missing because we couldn't stop them", etc...

Jake-Shadle: > BTW if we discuss making all the non-necessary sections fallible it would be nice to introduce a mechanism to gather multiple errors. Right now we know what will cause a failure but it's an all-or-nothing situation. It would be nice to have a mechanism so that we get a list of errors to distinguish between scenarios like "we could write the minidump but we don't have the file descriptors" or "we could only partially write a minidump, threads X, Y & Z are missing because we couldn't stop them", etc...

Totally agree, I'll open up a separate issue for this....oh wait, I already did and forgot about it #31.

gabrielesvelto:
@Jake-Shadle I've bumped a couple of dependencies to align this code to rust-minidump (which in turn was aligned with symbolic). If you're OK with those changes I'll merge this.

PR #98, opened 2024-01-16
Update memmap2 to 0.9

glandium:
Note this contains #97 because github sucks.


glandium:
Gah, this depends on https://github.com/rust-minidump/rust-minidump/pull/932

PR #99, opened 2024-02-05
update minidump-common dependencies

demoray:
Note: scroll must be upgraded due to it's shared dependency in minidump-common

Closes: #98


Jake-Shadle (approved):
Thanks!

PR #108, opened 2024-03-20
Send SIGSTOP to linux/android processes before doing other procfs/ptrace things.

### LINKED ISSUE #28: Improve the mechanism used to suspend threads
Suspending the target process threads is currently an all-or-nothing operation that prevents minidump generation [upon failure](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/minidump_writer.rs#L107). The system we currently use is to iterate over the threads we've enumerated [during initialization](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/ptrace_dumper.rs#L71). This is an inherently racy operation as new threads can be created in the meantime and existing ones can quit. This is partially handled in that we tolerate being unable to suspend certain threads assuming that we stop [at least a single one](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/ptrace_dumper.rs#L171-L172).

There should be ways to improve this but they're not clear-cut and probably needs some experimentation. One possibility is to send a `SIGSTOP` signal to the target process and wait for it to settle down before `ptrace()`-ing it. `SIGSTOP` is a big hammer that needs to be used with caution so I don't know what kind of issues we might encounter, but it could make the process more reliable by getting rid of some races (or at least leaving the kernel to handle them).

Additionally we might consider dumping threads we couldn't suspend. The data we'll get *might* be garbage but I don't really mind, we get garbage all the time in crash reports, it could also be *useful*. Let's say it would certainly be better than *nothing*. The reason why we never tried to do it with Breakpad is simple enough: handling errors is a chore! But here it's a lot easier. We could factorize the code for writing a single thread in `thread_list_stream::write()` and ignore individual failures (or even better, record them!).

afranchuk:
If this fails, we continue as we used to. This is an attempt to get a consistent/static process state.

Closes #28.

Needs to be tested in all sorts of scenarios (but due to the timeout and the fact that it continues on failure, one would hope it'd only be an improvement).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1226,6 +1226,7 @@ dependencies = [
  "futures",
  "goblin 0.8.0",
  "libc",
+ "log",
  "mach2",
  "memmap2",
  "memoffset",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,6 +264,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
+
+[[package]]
 name = "circular"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


Jake-Shadle (approved):

gabrielesvelto:
I had a couple of comments about this but wasn't fast enough, I'll put them in a separate issue.

afranchuk:
Yeah later last night I also realized that `process_stopped` should be set regardless of whether waiting for the status change succeeds.. (i.e. that needs to be fixed)

Jake-Shadle:
Oh right, I thought about something like starting if the bool was true or checking of the process has reached a stopped state in the interim, but I assume that sigcont is idempotent ?

afranchuk:
It also might not hurt to just always send SIGCONT, but it'd be cleaner to play it safe. I think as long as the SIGSTOP goes out, there's no reason to think the process won't stop (it's unblockable after all) even if we don't observe that.

gabrielesvelto:
Given the processes will frequently be in the same process group it might be possible to `waitpid(..., WUNTRACED)` the stopped process instead of polling, and fallback on polling if that doesn't work.

gabrielesvelto:
> Do you want to add yourself to https://github.com/rust-minidump/minidump-writer/blob/main/.github/CODEOWNERS so that you get auto-added as a reviewer? Then you can just remove the request if you don't have time/etc.

Yes, thank you!

Issue #79, opened 2023-05-03
Support libraries mapped from within APKs on Android


gabrielesvelto:
We've started encountering minidumps coming from Android builds of Firefox where the shared libraries have been mapped directly from an APK (which is just a ZIP file, so they're probably stored uncompressed). The relevant mappings look like this:
```
7e042700c000-7e042700e000 r--p 00001000 fd:20 7315                       /data/app/org.mozilla.geckoview.test_runner-1/split_config.x86_64.apk
7e042700f000-7e042707e000 r-xp 00002000 fd:20 7315                       /data/app/org.mozilla.geckoview.test_runner-1/split_config.x86_64.apk
7e042707e000-7e04270af000 r--p 00071000 fd:20 7315                       /data/app/org.mozilla.geckoview.test_runner-1/split_config.x86_64.apk
7e04270b0000-7e04270b2000 r--p 000a2000 fd:20 7315                       /data/app/org.mozilla.geckoview.test_runner-1/split_config.x86_64.apk
7e04270b2000-7e04270b3000 rw-p 000a4000 fd:20 7315                       /data/app/org.mozilla.geckoview.test_runner-1/split_config.x86_64.apk
```
Since we record the module name we find in the mappings the minidumps are unusable, as we can't symbolicate them.

We look for the SONAME in the mapped [ELF file](https://github.com/rust-minidump/minidump-writer/blob/a73769022d585ecf42c7cec17039d71b923b39ad/src/linux/maps_reader.rs#L286) so we should be finding it, but chances are that we are failing to open the underlying file [here](https://github.com/rust-minidump/minidump-writer/blob/a73769022d585ecf42c7cec17039d71b923b39ad/src/linux/maps_reader.rs#L267).

So this could be related to issue https://github.com/rust-minidump/minidump-writer/issues/71, because we'd be certainly capable of finding the SONAME within the crashed process' memory. If we switch reading the build ID from there, reading the SONAME would be a trivial addition.

PR #112, opened 2024-03-29
Read ELF build ids directly from the target process.

### LINKED ISSUE #71: Read ELF build ids directly from the target process instead of mmap()ing it
When populating the module list on Linux we extract the GNU build ID from each executable. To do so we first `mmap()` the entire file in the process writing the minidump and then extract the data from there. This has a couple of important drawbacks:
- If the file on disk has been deleted we'll get an empty ID, something that we see often in Firefox crashes.
- If the file on disk has been altered we'll get the wrong ID.
- If the file is extremely large `mmap()` may fail on 32-bit hosts, we've also seen this happen in 32-bit builds of Firefox.

We could avoid all these issues by reading the ELF headers directly from the process we're dumping. `goblin` supports parsing an ELF file lazily, though one has to do it manually (see [this example](https://github.com/m4b/goblin/blob/4c99ae8f3910d66a20b640857646d10827042460/tests/elf.rs#L37-L76)).

afranchuk:
Closes #71.

A few things to consider:
* Since we read from the process memory, the process must be in ptrace-stop (see `test_file_id`). This changes when the build ids can be read. Previously they could be read without the process being stopped if the mapped files still existed (and were hopefully the same that the process was using).
* The previous implementation made some mutations to deleted mapping names (removing the ` (deleted)` suffix). We need to decide whether we still want/need this behavior. In the meantime I commented out a failing test assertion.


afranchuk:
Hmm, not sure why the process parent in `test_file_id` seems to be gone in CI (I can't reproduce this locally).

src/linux/build_id_reader.rs
```diff
@@ -0,0 +1,225 @@
+use crate::errors::BuildIdReaderError as Error;
+use crate::minidump_format::GUID;
+use goblin::{
+    container::{Container, Ctx, Endian},
+    elf,
+};
+
+const NOTE_SECTION_NAME: &[u8] = b".note.gnu.build-id\0";
+
+pub trait ModuleMemory {
```
Jake-Shadle, line 10: Not sure what the point of using a trait here is if there is only one implementor

Jake-Shadle (changes requested):

src/linux/build_id_reader.rs
afranchuk: There are two. https://github.com/rust-minidump/minidump-writer/pull/112/files#diff-ed2349064f78dacbfa376d37df89b784688730cc1a397858b847cc5399f142a9R584-R600

src/linux/ptrace_dumper.rs
```diff
@@ -634,39 +567,44 @@ impl PtraceDumper {
...
-                return Self::elf_file_identifier_from_mapped_file(mem_slice);
-            } else {
-                let mem_slice = Self::copy_from_process(
-                    pid,
-                    mapping.start_address as *mut libc::c_void,
-                    mapping.size,
-                )?;
-                return Self::elf_file_identifier_from_mapped_file(&mem_slice);
+        // FIXME `fixup_deleted_file` mutations of `mapping` were removed. Will this have a
+        // meaningful effect elsewhere?
```
gabrielesvelto, line 571: If the file was moved then the new code should be able to read from it and the fixup shouldn't be needed anymore. Similarly, if it's been deleted the kernel should have been holding on to its copy in memory. My guess is that we don't need to cover any of these cases explicitly anymore.

gabrielesvelto (changes requested):

src/linux/build_id_reader.rs
```diff
@@ -0,0 +1,225 @@
...
+    module_memory: T,
+    header: elf::Header,
+    context: Ctx,
+}
+
+impl<T: ModuleMemory> ElfBuildIdReader<T> {
+    pub fn new(module_memory: T) -> Result<Self, Error> {
+        // We could use `Ctx::default()` (which defaults to the native system), however to be extra
+        // permissive we'll just use a 64-bit ("Big") context which would result in the largest
+        // possible header size.
```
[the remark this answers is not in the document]
afranchuk: The only not-completely-contrived case I could think of is the minidump-writer process being 64-bit and the child process being 32-bit, or vice versa. But I think it _is_ very unlikely, especially nowadays.

src/linux/build_id_reader.rs
```diff
@@ -0,0 +1,225 @@
...
+            &self.module_memory,
+            self.header.e_shoff,
+            self.header.e_shentsize as u64 * self.header.e_shnum as u64,
+        )?;
+        let section_headers = elf::SectionHeader::parse(
+            &section_headers_data,
+            0,
+            self.header.e_shnum as usize,
+            self.context,
+        )?;
```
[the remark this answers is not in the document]
afranchuk: I actually almost did that and cached the `section_headers`. I'll probably do that; I didn't because I thought the likelihood of not having a build id was so low it didn't matter much anyway (and I wanted to keep these methods `&self` rather than `&mut self`).

src/linux/ptrace_dumper.rs
afranchuk: I think maybe we still want to remove ` (deleted)` from names though? Otherwise the writer will write e.g. the main module with that suffix (as in the `with_deleted_binary` test where I commented out the assertion), which isn't what we want.

src/linux/build_id_reader.rs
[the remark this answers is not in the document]
gabrielesvelto: Ah yes, because we're reading it from `/proc/<pid>/maps` so we need to strip the suffix for deleted files (or it'd mess up with symbol resolution later when we process the minidump). Somehow my mind had conflated the fact that we also want to read the **SONAME** from memory with this patch, but that's actually issue #79 which is the natural follow-up to this.

gabrielesvelto (changes requested):
I've left a small comment but noticed a bigger issue. On my Gentoo machine there's a lot of failures and I'm not 100% sure what's going on, however what I see consistently is the following:
- We call `PtraceDumper::elf_identifier_for_mapping()` on `/usr/lib/gcc/x86_64-pc-linux-gnu/13/libgcc_s.so.1` which in turns calls `build_id_reader::read_build_id()`
- `build_id_reader::read_build_id()` tries `ElfBuildIdReader::read_from_program_headers()` which fails, because this file doesn't have a build ID so we can't find the note
- `build_id_reader::read_build_id()` proceeds with `ElfBuildIdReader::read_from_section()` and here things go south. `read_section_headers()` calls `elf::SectionHeader::parse()` with an offset of 0 which causes it to return an empty array, see [here](https://github.com/m4b/goblin/blob/47ee850af2beaf81ed5d99b5546de75db8e33b75/src/elf/section_header.rs#L448-L451)
- We attempt to execute this on an empty array and the test fails
  ```rustlang
  let strtab_section_header = &section_headers[self.header.e_shstrndx as usize];
  ```
One very crude way of fixing this is to add 1 byte in front of the slice containing the section headers, so that goblin really parses them instead of returning the empty vector. That being said we should still handle the case where `self.header.e_shstrndx` ends up being out-of-bounds, even though it might have masked this failure.

afranchuk:
Oh no! Looks like we could just use `parse_from()` once the new version of goblin is published with @lissyx's changes.

afranchuk:
It seems like some more unit tests to cover these cases are necessary, though it will be a bit contrived to e.g. create binaries with only a build id in them (to not waste space in the repo). But I'll try to add something.

afranchuk:
I've added tests (with a small, handcrafted ELF file). I made the ELF file by making a bunch of modifications to https://github.com/tchajed/minimal-elf; I'm wondering whether I should include the code that created it in the repo?

I sadly just realized that goblin does implement `Pwrite` (just not for the 32/64-bit agnostic `Elf` wrapper which I originally was looking at), so I could/should probably rewrite the generation code using goblin. In that case keeping the size as small as possible isn't a big deal since it will be generated at runtime, but it'll likely end up the same size anyway since I don't think goblin will add anything we don't explicitly tell it to.

gabrielesvelto:
After a bit of wrangling with the `with_deleted_binary` test I figured out why it's failing on my machine, and it's not such a big deal. Locally the test program doesn't have the build ID program note and the section headers aren't being loaded into memory, or the kernel has discarded that particular area after we deleted the file. So what happens is the following:
* `read_build_id()` will attempt to find the note in the program headers and fail (it's not there)
* `read_build_id()` will fall back onto `ElfBuildIdReader::generate_from_section()` but this fails, as the section headers are not in memory
* `read_build_id()` will fall back onto `ElfBuildIdReader::generate_from_text()`, but this needs the section header to find the text section, so this also fails

Given that I don't know a way to recover the text section from the program headers alone, and that the failure is triggered by an executable that doesn't have a build ID in the first place I think we can let it slide. We'll never encounter this in practice within Firefox, because the program headers of our various executables will always include the build ID.

gabrielesvelto (approved):
This is OK as it is.

Jake-Shadle (approved):

PR #115, opened 2024-04-10
Add android testing CI job

afranchuk:
Creating a PR to run and develop the job.

### DIFF
--- a/.cargo/android-runner.sh
+++ b/.cargo/android-runner.sh
@@ -1,0 +1,9 @@
+#!/bin/sh -e
+
+BINARY=$1
+shift
+
+# Make sure to run the following to copy the test helper binary over.
+# cargo run --target x86_64-linux-android --bin test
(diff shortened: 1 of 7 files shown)


gabrielesvelto:
I'm going to defer this review to @Jake-Shadle as my knowledge of the GitHub CI machinery is very limited.

Jake-Shadle (approved):
LGTM, and doesn't seem to be wildly slower than the other jobs, good to finally have this tested, thanks! Is there the possibility of targeting aarch64 android as well? While testing that the syscalls on android work as expected, there is still a gap of any code that happens to be specific to both android + aarch64 that we could miss.

afranchuk:
@Jake-Shadle yes, I believe we can! I'll try that out as well. FWIW there's also a way to cache the android virtual device, I'll try that out to see how much faster it is.

PR #116, opened 2024-04-11
Add aarch64-android testing to CI

afranchuk:
Let's see if this works... I didn't try locally testing aarch64-android because I didn't want to jump through hoops to install the aarch64 qemu and get that working (Google dropped support so it doesn't Just Work).

### DIFF
--- a/.cargo/android-runner.sh
+++ b/.cargo/android-runner.sh
@@ -3,7 +3,12 @@
 BINARY=$1
 shift
 
+# When pushing a bin, the full target dir is appended (including the target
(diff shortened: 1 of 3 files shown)


afranchuk:
Due to https://github.com/ReactiveCircus/android-emulator-runner/issues/350, the only aarch64(ish) github actions platform, `macos-14`, doesn't work. We can't test `aarch64-android` without an `aarch64` runner.

Jake-Shadle:
Well that's extremely unfortunate.

PR #117, opened 2024-04-19
Read ELF SONAMEs directly from process memory when possible.

### LINKED ISSUE #79: Support libraries mapped from within APKs on Android

afranchuk:
When this fails, it falls back to the previous behavior of trying to open the mapped file name directly.

Closes #79.


afranchuk:
I'll update the ELF fixture to add an SONAME to test this.

Jake-Shadle (approved):

gabrielesvelto:
Sorry my review is taking a bit but I'm doing some testing directly in Firefox and this seems to work only for external libraries, not our own ones, so I'm trying to figure out what's wrong (or if it's something specific to my box).

src/linux/module_reader.rs
```diff
@@ -0,0 +1,433 @@
...
+        let context = Ctx::new(header.container()?, header.endianness()?);
+        Ok(ModuleReader {
+            module_memory,
+            header,
+            context,
+        })
+    }
+
+    pub fn soname(&self) -> Result<String, Error> {
+        let section_headers = self.read_section_headers()?;
```
gabrielesvelto, line 162: For some reason this call is always failing when encountering our executables in Firefox, even though it works fine for system libraries. Could you check with one of our builds and see what's going on?

Note that I tried this with an elf-hack build because that's what we ship.

gabrielesvelto (changes requested):

src/linux/module_reader.rs
afranchuk: Interesting, I'll try it out!

afranchuk: I'm unable to reproduce this, both with firefox release and nightly (as downloaded directly from our website). Is there something else I need to do? Can you tell me exactly what you did to test? I added a test like
```rust
    #[test]
    fn soname_firefox() {
        let bytes = include_bytes!("/home/afranchuk/Downloads/125/firefox/libxul.so");
        let reader = ModuleReader::new(&bytes[..]).unwrap();
        let soname = reader.soname().unwrap();
        assert_eq!(soname, "libxul.so");
    }
```

I tried with a few other libraries too.

gabrielesvelto: I patched the gecko sources to build minidump-writer with this PR included and crashed a child process to check how the code would behave in the wild.

afranchuk: Perhaps this is a vaddr vs file offset issue. I'll try that.

afranchuk:
Not all binaries have the section header table present when loaded in memory. However the program headers will be there, so I added another path which is preferred and uses only the program headers and the dynamic section information. This seems to work with all shared objects loaded by firefox at runtime when I tested it. We _might_ still want a fallback of trying to load directly from the file (if present) if reading from memory fails for some reason.

afranchuk:
I'm not sure why the android test is failing...

.cargo/config.toml
```diff
@@ -11,4 +11,6 @@
 # ]
 [target.x86_64-linux-android]
 linker = "x86_64-linux-android30-clang"
+# By default the linker _doesn't_ generate a build-id, however we want one for our tests.
+rustflags = ["-C", "link-args=-Wl,--build-id=sha1"]
```
gabrielesvelto, line 15: Nice catch!

afranchuk:
This is probably ready to merge.

gabrielesvelto (approved):
Indeed, LGTM

PR #125, opened 2024-07-08
update minidump dependencies

demoray:


demoray:
As an aside, it would be useful to have the minidump-writer release cycle aligned in the minidump crate releases.  Is this something this project looks to achieve eventually?

gabrielesvelto:
Thanks, I've merged these changes in PR #126 which also fixes the Android automation failure.

PR #126, opened 2024-07-09
Update the minidump crates to 0.22

afranchuk:
This creates a few duplicate dependencies, but they are only dev-dependencies (and they will work themselves out as the `backtrace` crate gets updated).

I also realized some of the ptrace_dumper tests fail on Android for (likely) the same reason they fail in CI, so I have a commit addressing that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,12 +4,21 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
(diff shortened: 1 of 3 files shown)


gabrielesvelto (approved):
LGTM

afranchuk:
@Jake-Shadle can I get a new release that includes this PR?

PR #127, opened 2024-07-16
Setup infrastructure to allow Auxv information to be passed from crashed process.

### LINKED ISSUE #27: Improve handling of the auxiliary vector on Linux
Right now if we can't read the auxiliary vector by parsing the corresponding file under /proc [minidump generation fails](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/ptrace_dumper.rs#L70). This is a known cause of failures to write minidumps on Android (see [bug 1644486](https://bugzilla.mozilla.org/show_bug.cgi?id=1644486) on our tracker for example) and Linux because sometimes the generator process doesn't have permission to access that file. This accounts for at least one out of three failures when writing Linux/Android minidumps.

We currently use the auxiliary vector in three different places:
* To get the location of [linux-gate.so](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/ptrace_dumper.rs#L252)
* To get the [entry point](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/ptrace_dumper.rs#L266)
* To write the [DSO debug stream](https://github.com/rust-minidump/minidump-writer/blob/d34d50fac678239af8e662fb636166c28fd3ae1e/src/linux/dso_debug.rs#L80)

Neither of those uses is strictly needed. In fact the first two instance already handle the case where the relevant entry in the auxiliary vector isn't available, and the third one is just not critical. We could just skip writing the DSO stream and everybody would be still perfectly happy.

So I propose to do a couple of things to improve this situation:
* Make not being able to read the auxiliary vector a non-fatal failure (this is easy)
* Change how it's read to have a fallback in case we can't access `/proc`. Notably the auxiliary vector is [at a known location in memory](https://lwn.net/Articles/519085/) so we could pull it out of there using `ptrace()`.

marti4d:
If the crashed process supports it, it can directly pass in the needed values from the auxiliary vector to avoid potential issues where "/proc/{pid}/auxv" can't be accessed on Android.

If the information is not passed in (or is incomplete), the code will attempt to fill in any gaps from the auxv file. If any of this fails, it will now just log a warning message and continue writing the minidump.

Fixes #27

### DIFF
--- a/src/linux/auxv/mod.rs
+++ b/src/linux/auxv/mod.rs
@@ -14,20 +14,14 @@ pub type AuxvType = u32;
 #[cfg(target_pointer_width = "64")]
 pub type AuxvType = u64;
 
-#[cfg(any(target_arch = "arm", all(target_os = "android", target_arch = "x86")))]
-mod consts {
-    use super::AuxvType;
-    pub const AT_PHDR: AuxvType = 3;
-    pub const AT_PHNUM: AuxvType = 5;
-    pub const AT_ENTRY: AuxvType = 9;
-    pub const AT_SYSINFO_EHDR: AuxvType = 33;
-}
-#[cfg(not(any(target_arch = "arm", all(target_os = "android", target_arch = "x86"))))]
 mod consts {
     use super::AuxvType;
     pub const AT_PHDR: AuxvType = libc::AT_PHDR;
     pub const AT_PHNUM: AuxvType = libc::AT_PHNUM;
     pub const AT_ENTRY: AuxvType = libc::AT_ENTRY;
+    #[cfg(target_os = "android")]
+    pub const AT_SYSINFO_EHDR: AuxvType = 33;
+    #[cfg(not(target_os = "android"))]
     pub const AT_SYSINFO_EHDR: AuxvType = libc::AT_SYSINFO_EHDR;
 }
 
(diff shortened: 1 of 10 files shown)


src/linux/auxv/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        if self.linux_gate_address.is_none() {
+            self.linux_gate_address = auxv.get(&consts::AT_SYSINFO_EHDR).copied();
+        }
+
+        if self.entry_address.is_none() {
+            self.entry_address = auxv.get(&consts::AT_ENTRY).copied();
+        }
+
```
Jake-Shadle, line 102: Seems wasteful to do the temporary hashmap heap allocations just to pull out a few bits of data, seems simpler to just pull them out while iterating, which would also allow early out if all items we're interested in are set.

Jake-Shadle: ...though now realizing this was just moved code.. :stuck_out_tongue:

Jake-Shadle (changes requested):

src/linux/auxv/mod.rs
marti4d: Yep -- It was moved code... But it's still a good point! I guess moving the code just made it more obvious how un-needed the `HashMap` was. I changed it :)

gabrielesvelto (approved):
LGTM too, @marti4d is this ready for merging?

marti4d:
@gabrielesvelto It is! And the sooner, the better, since I'm going to have to get this vendored into M-C so the C++ part of my change can use it :)

PR #129, opened 2024-08-07
Perform a checked add in the ptrace dumper ModuleMemory impl.

afranchuk:
Without this, integer overflow will cause a panic when we should gracefully fail. It's very possible the start address or offset could be incorrect/corrupted.

### DIFF
--- a/src/linux/module_reader.rs
+++ b/src/linux/module_reader.rs
@@ -126,7 +126,11 @@ fn section_header_with_name<'sc>(
     name: &[u8],
     module_memory: &mut ProcessMemory<'_>,
 ) -> Result<Option<&'sc elf::SectionHeader>, Error> {
-    let strtab_section_header = section_headers.get(strtab_index).ok_or(Error::NoStrTab)?;
+    let strtab_section_header = section_headers
+        .get(strtab_index)
+        .and_then(|hdr| (hdr.sh_type == elf::section_header::SHT_STRTAB).then_some(hdr))
+        .ok_or(Error::NoStrTab)?;
+
     for header in section_headers {
         let sh_name = header.sh_name as u64;
         if sh_name >= strtab_section_header.sh_size {
--- a/src/linux/module_reader.rs
+++ b/src/linux/module_reader.rs
@@ -45,32 +45,36 @@ impl<'buf> From<ProcessReader> for ProcessMemory<'buf> {
 impl<'buf> ProcessMemory<'buf> {
     #[inline]
     fn read(&mut self, offset: u64, length: u64) -> Result<Buf<'buf>, Error> {
+        let error = move |start_address, error| Error::ReadModuleMemory {
+            start_address,
+            offset,
+            length,
+            error,
+        };
+
         match self {
             Self::Process(pr) => {
-                let len = std::num::NonZero::new(length as usize).ok_or_else(|| {
-                    Error::ReadModuleMemory {
-                        offset,
-                        length,
-                        error: nix::errno::Errno::EINVAL,
-                    }
-                })?;
+                let error = |e| error(Some(pr.start_address), e);
+                let len = std::num::NonZero::new(length as usize)
+                    .ok_or_else(|| error(nix::Error::EINVAL))?;
+                let proc_offset = pr
+                    .start_address
+                    .checked_add(offset)
+                    .ok_or_else(|| error(nix::Error::EOVERFLOW))?;
                 pr.inner
-                    .read_to_vec((pr.start_address + offset) as _, len)
+                    .read_to_vec(proc_offset as _, len)
                     .map(Cow::Owned)
-                    .map_err(|err| Error::ReadModuleMemory {
-                        offset,
-                        length,
-                        error: err.source,
-                    })
+                    .map_err(|err| error(err.source))
             }
-            Self::Slice(s) => s
-                .get(offset as usize..(offset + length) as usize)
-                .map(Cow::Borrowed)
-                .ok_or_else(|| Error::ReadModuleMemory {
-                    offset,
-                    length,
-                    error: nix::Error::EACCES,
-                }),
+            Self::Slice(s) => {
+                let error = |e| error(None, e);
+                let end = offset
+                    .checked_add(length)
+                    .ok_or_else(|| error(nix::Error::EOVERFLOW))?;
+                s.get(offset as usize..end as usize)
+                    .map(Cow::Borrowed)
+                    .ok_or_else(|| error(nix::Error::EACCES))
+            }
         }
     }
 
@@ -141,7 +145,7 @@ fn section_header_with_name<'sc>(
     Ok(None)
 }
 
-/// Types which can be read from an `impl ModuleMemory`.
+/// Types which can be read from ProcessMemory.
 pub trait ReadFromModule: Sized {
     fn read_from_module(module_memory: ProcessMemory<'_>) -> Result<Self, Error>;
 }
@@ -279,7 +283,12 @@ impl<'buf> ModuleReader<'buf> {
             (_, _, None) => Err(Error::NoSoNameEntry),
             (Some(addr), Some(size), Some(offset)) => {
                 // If loaded in memory, the address will be altered to be absolute.
-                self.read_name_from_strtab(self.module_memory.absolute(addr), size, offset)
+                if offset < size {
+                    self.read_name_from_strtab(self.module_memory.absolute(addr), size, offset)
+                } else {
+                    log::warn!("soname strtab offset ({offset}) exceeds strtab size ({size})");
+                    Err(Error::NoSoNameEntry)
+                }
             }
         }
     }
@@ -319,6 +328,11 @@ impl<'buf> ModuleReader<'buf> {
                         self.section_offset(dynstr_section_header),
                         dynstr_section_header.sh_size,
                         name_offset,
+                    );
+                } else {
+                    log::warn!(
+                        "soname offset ({name_offset}) exceeds dynstr section size ({})",
+                        dynstr_section_header.sh_size
                     );
                 }
             }
@@ -394,6 +408,7 @@ impl<'buf> ModuleReader<'buf> {
         strtab_size: u64,
         name_offset: u64,
     ) -> Result<String, Error> {
+        assert!(name_offset < strtab_size);
         let name = self
             .module_memory
             .read(strtab_offset + name_offset, strtab_size - name_offset)?;
@@ -436,6 +451,7 @@ impl<'buf> ModuleReader<'buf> {
             self.header.e_shoff,
             self.header.e_shentsize as u64 * self.header.e_shnum as u64,
         )?;
+        // Use `parse_from` rather than `parse`, which allows a 0 offset.
         let section_headers = elf::SectionHeader::parse_from(
             &section_headers_data,
             0,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ thiserror = "1.0"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
-goblin = "0.8"
+goblin = "0.8.2"
 memmap2 = "0.9"
 
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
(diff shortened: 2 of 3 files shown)


afranchuk:
I'm augmenting this with a bunch of other arithmetic checks throughout the module_reader code.

mstange:
Do you have the /proc/pid/maps from when you saw the incorrect / corrupted state? I wonder if it's not corruption but something else going on.

afranchuk:
I do not have the inputs, though I would like to have them. It was on Android in CI testing: https://treeherder.mozilla.org/logviewer?job_id=469290206&repo=autoland&lineNumber=8170.

gabrielesvelto:
This is bizarre and might be hiding a deeper problem. Is the issue reproducible? Can you check what values are involved and where did they come from?

afranchuk:
I was going to try some runs with debug logging to try to track it down.

afranchuk:
Regardless of the underlying issue, we _are_ doing arithmetic with values derived from inputs, so I think we should not panic if overflow occurs.

gabrielesvelto:
> Regardless of the underlying issue, we _are_ doing arithmetic with values derived from inputs, so I think we should not panic if overflow occurs.

Yes, I agree.

afranchuk:
The problem is reproducible, now I'm trying to figure out how to exfiltrate some debug information.

gabrielesvelto:
> The problem is reproducible, now I'm trying to figure out how to exfiltrate some debug information.

`eprintln!()` FTW

afranchuk:
`eprintln!()` did not work!

afranchuk:
I've still been unable to figure out why this fails, however the fix here does remedy the error. While I'd like to know the root cause, it'd be nice to get this merged.

gabrielesvelto:
I just re-checked the failed workload and I think there might be a way to surface the contents of the memory map. You could try using the `android_log-sys` or `android-logger` crate we have vendored to dump the contents of the memory map directly to the logcat. Our automation tasks capture the entire logcat output, so we could try that to exfiltrate the information we need.

afranchuk:
I'm not sure whether `android-logger` will work, as I had assumed that when linking it links to the same log crate instance in gkrust (which presumably has a logger registered). However, it's entirely possible that's not happening, so I'm going to try it out.

gabrielesvelto:
Worst case `android_log-sys` should allow you to invoke `__android_log_print()` directly, that will surely end up in the logcat output.

afranchuk:
`android-logger` didn't work; it's documentation says that if a logger is already installed it will silently do nothing, so I can only guess that's what's going on. I'll try directly printing.

afranchuk:
Okay, so it looks like `android-logger` and `__android_log_print()` are both working, which means `android-logger` was working before and the taskcluster UI for searching the log is broken (I downloaded the log this time when I still didn't see the logs I expected).

afranchuk:
The overflow occurs here:
```
start address=129746557718528, offset=18446738911158816776
```
This offset occurs when trying to get the build id of `/system/lib64/libGLESv2.so` from the notes section. Based on the addresses being read according to the log and the crash backtrace, I believe the strtab section offset is being read incorrectly, though I don't see any obvious reasons for this. However we can check this offset against the known size of the module. I'm going to try to get my hands on that file to confirm this.

afranchuk:
The file works as expected when our code runs on it directly, so something odd is happening when it is loaded in memory. I noticed that the section header offset _in the file_ is `0x16170`, however according to the logs the offset _during the failing test_ was `0x1616f`, which I find a bit suspicious (though it's possible that's how it's loaded, but even the size in memory is the same as the file size). I can dump the loaded module to see what's up there. I also can't find that value in the file (if the section offset was being misread).

gabrielesvelto:
Could you try checking if the issue presents itself with elfhack disabled?

afranchuk:
It appears as if the section header table is not loaded. This is not surprising, I've seen this before, though it's uncommon. I'll double-check the program headers to make sure this is expected. So in this case, there's two things we can do: verify the type of section headers we read/use (it's likely to be junk) before doing anything with them, and do checked arithmetic to avoid garbage values (which is already implemented).

afranchuk:
The program headers confirm that the section header table wasn't expected to be loaded. What's disappointing is that the program header PT_NOTE section doesn't contain the build id (which is what we're trying to get), only the section has it. The section is loaded into memory: it happens to be right after the PT_NOTE data.

mstange:
Can you paste the program header table and the section table here? (I mean the output of readelf that displays those tables)
Which program header is the section table in? How is it different in the normal case?

afranchuk:
> Can you paste the program header table and the section table here? (I mean the output of readelf that displays those tables) Which program header is the section table in? How is it different in the normal case?

```
There are 25 section headers, starting at offset 0x16170:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .interp           PROGBITS         0000000000000238  00000238
       0000000000000015  0000000000000000   A       0     0     1
  [ 2] .note.androi[...] NOTE             000000000000024e  0000024e
       0000000000000018  0000000000000000   A       0     0     2
  [ 3] .note.gnu.bu[...] NOTE             0000000000000268  00000268
       0000000000000020  0000000000000000   A       0     0     4
  [ 4] .dynsym           DYNSYM           0000000000000288  00000288
       00000000000044d0  0000000000000018   A       5     1     8
  [ 5] .dynstr           STRTAB           0000000000004758  00004758
       0000000000003ccb  0000000000000000   A       0     0     1
  [ 6] .gnu.hash         GNU_HASH         0000000000008428  00008428
       0000000000001590  0000000000000000   A       4     0     8
  [ 7] .gnu.version      VERSYM           00000000000099b8  000099b8
       00000000000005bc  0000000000000002   A       4     0     2
  [ 8] .gnu.version_d    VERDEF           0000000000009f74  00009f74
       000000000000001c  0000000000000000   A       5     1     4
  [ 9] .gnu.version_r    VERNEED          0000000000009f90  00009f90
       0000000000000020  0000000000000000   A       5     1     4
  [10] .rela.dyn         RELA             0000000000009fb0  00009fb0
       0000000000000030  0000000000000018   A       4     0     8
  [11] .rela.plt         RELA             0000000000009fe0  00009fe0
       0000000000000090  0000000000000018  AI       4    12     8
  [12] .plt              PROGBITS         000000000000a070  0000a070
       0000000000000070  0000000000000010  AX       0     0     16
  [13] .text             PROGBITS         000000000000a0e0  0000a0e0
       0000000000005c45  0000000000000000  AX       0     0     16
  [14] .eh_frame         PROGBITS         000000000000fd28  0000fd28
       00000000000044f4  0000000000000000   A       0     0     8
  [15] .eh_frame_hdr     PROGBITS         000000000001421c  0001421c
       00000000000016d4  0000000000000000   A       0     0     4
  [16] .fini_array       FINI_ARRAY       0000000000016d60  00015d60
       0000000000000008  0000000000000000  WA       0     0     8
  [17] .dynamic          DYNAMIC          0000000000016d68  00015d68
       0000000000000250  0000000000000010  WA       5     0     8
  [18] .got              PROGBITS         0000000000016fb8  00015fb8
       0000000000000000  0000000000000000  WA       0     0     8
  [19] .got.plt          PROGBITS         0000000000016fb8  00015fb8
       0000000000000048  0000000000000000  WA       0     0     8
  [20] .data             PROGBITS         0000000000017000  00016000
       0000000000000008  0000000000000000  WA       0     0     8
  [21] .comment          PROGBITS         0000000000000000  00016008
       0000000000000026  0000000000000001  MS       0     0     1
  [22] .note.gnu.go[...] NOTE             0000000000000000  00016030
       000000000000001c  0000000000000000           0     0     4
  [23] .gnu_debuglink    PROGBITS         0000000000000000  0001604c
       0000000000000014  0000000000000000           0     0     1
  [24] .shstrtab         STRTAB           0000000000000000  00016060
       0000000000000109  0000000000000000           0     0     1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  D (mbind), l (large), p (processor specific)

Elf file type is DYN (Shared object file)
Entry point 0x0
There are 9 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000000040 0x0000000000000040
                 0x00000000000001f8 0x00000000000001f8  R      0x8
  INTERP         0x0000000000000238 0x0000000000000238 0x0000000000000238
                 0x0000000000000015 0x0000000000000015  R      0x1
      [Requesting program interpreter: /system/bin/linker64]
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x00000000000158f0 0x00000000000158f0  R E    0x1000
  LOAD           0x0000000000015d60 0x0000000000016d60 0x0000000000016d60
                 0x00000000000002a8 0x00000000000002a8  RW     0x1000
  DYNAMIC        0x0000000000015d68 0x0000000000016d68 0x0000000000016d68
                 0x0000000000000250 0x0000000000000250  RW     0x8
  NOTE           0x000000000000024e 0x000000000000024e 0x000000000000024e
                 0x0000000000000018 0x0000000000000018  R      0x4
  GNU_EH_FRAME   0x000000000001421c 0x000000000001421c 0x000000000001421c
                 0x00000000000016d4 0x00000000000016d4  R      0x4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x0
  GNU_RELRO      0x0000000000015d60 0x0000000000016d60 0x0000000000016d60
                 0x00000000000002a0 0x00000000000002a0  RW     0x8

 Section to Segment mapping:
  Segment Sections...
   00     
   01     .interp 
   02     .interp .note.android.ident .note.gnu.build-id .dynsym .dynstr .gnu.hash .gnu.version .gnu.version_d .gnu.version_r .rela.dyn .rela.plt .plt .text .eh_frame .eh_frame_hdr 
   03     .fini_array .dynamic .got .got.plt .data 
   04     .dynamic 
   05     .note.android.ident 
   06     .eh_frame_hdr 
   07     
   08     .fini_array .dynamic .got .got.plt 
```

The section header table is not in a program header (which is the problem). The `.note.gnu.build-id` section is within the read/execute PT_LOAD header (header 2). Most binaries look like that, with an RX `PT_LOAD` from 0 to the end of the section contents which should be RX (they seem to be ordered intentionally to keep those sections all together).

It seems like this is the common case, I was wrong earlier to say that it's uncommon (though on linux I definitely recall having seen some binaries loaded wholly into memory, including the section header table). I think we normally succeed in getting the build id from the program headers, so we don't attempt the section headers (or we do but just get an access violation which we handle fine, rather than this overflow).

afranchuk:
Maybe if we can't get the build id by reading the in-memory binary we could try to read it from the file (if present)? We changed to reading in-memory because in some cases the files are not present, but it's not an unreasonable fallback.

src/linux/module_reader.rs
```diff
@@ -49,28 +49,48 @@ impl<'buf> ProcessMemory<'buf> {
             Self::Process(pr) => {
                 let len = std::num::NonZero::new(length as usize).ok_or_else(|| {
                     Error::ReadModuleMemory {
+                        start_address: Some(pr.start_address),
                         offset,
                         length,
                         error: nix::errno::Errno::EINVAL,
                     }
```
gabrielesvelto, line 56: Given the repetition of this rather large block it might be worth replacing it through the function with a closure, like:
```rust
let error = |start_address, error| {
    Error::ReadModuleMemory {
        start_address,
        offset,
        length,
        errror,
    }
}
```

gabrielesvelto (changes requested):
Just a small change to make the code leaner but the rest is good.

src/linux/module_reader.rs
afranchuk: That was bugging me too. Will do!

gabrielesvelto (approved):
LGTM

Issue #124, opened 2024-06-05
Adjust how we deal with pending signals when stopping threads on Linux


gabrielesvelto:
I was looking through [this patch](https://chromium.googlesource.com/breakpad/breakpad/+/527331a471207d28d015e210c12fb8bb8b6732fd%5E%21/#F0) in Google Breakpad, and it changes how they wait for a thread to be  stopped after attaching to it via `ptrace()`. In short, there might be a race between the `SIGSTOP` signal sent when suspending a thread and what we get in the following `waitpid()`. The different signals that might fall between the two syscalls can get lost, and the thread won't behave correctly once resumed, as it will eventually be stopped again.

PR #131, opened 2024-08-12
Handle non-SIGSTOP signals

Jake-Shadle:
- **Handle non-SIGSTOP signals**
- **Add failed attempt to test**

This just implements the code pointed out in #124. I attempted to add a test for this by starting another named thread and sending a SIGHUP to it before spawning the dumper, but I don't think this scenario will ever result in the new code being run.

Resolve: #124

### DIFF
--- a/src/linux/ptrace_dumper.rs
+++ b/src/linux/ptrace_dumper.rs
@@ -143,8 +143,31 @@ impl PtraceDumper {
         ptrace::attach(pid).map_err(|e| AttachErr(child, e))?;
         loop {
             match wait::waitpid(pid, Some(wait::WaitPidFlag::__WALL)) {
-                Ok(_) => break,
-                Err(_e @ Errno::EINTR) => continue,
+                Ok(status) => {
+                    let wait::WaitStatus::Stopped(_, status) = status else {
+                        return Err(DumperError::WaitPidError(
+                            child,
+                            nix::errno::Errno::UnknownErrno,
+                        ));
+                    };
+
+                    // Any signal will stop the thread, make sure it is SIGSTOP. Otherwise, this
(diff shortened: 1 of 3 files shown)


src/linux/ptrace_dumper.rs
```diff
@@ -160,8 +160,31 @@ impl PtraceDumper {
...
+                    // signal will be delivered after PTRACE_DETACH, and the thread will enter
+                    // the "T (stopped)" state.
+                    if status == nix::sys::signal::SIGSTOP {
+                        break;
+                    }
+
+                    // Signals other than SIGSTOP that are received need to be reinjected,
+                    // or they will otherwise get lost.
+                    // Note that the breakpad code doesn't do a detach, but that
+                    // feels like a bug...
```
gabrielesvelto, line 181: I was reading the `ptrace(2)` manpage and it turns out this is complicated to say the least. Detaching after the process woke up won't work because for it to work the thread must be stopped, so the page suggests re-stopping and re-injecting the signals again before detaching. So Breakpad is not wrong about not detaching but it's not doing the right thing either. I'm gonna check what gdb and lldb do to see if there's a way that doesn't require several nested loops of attaching/waiting/detaching to get it right. BTW the manpage seems to implicitly admit that this was a genuine design flaw.

gabrielesvelto (changes requested):

gabrielesvelto (approved):

gabrielesvelto:
LGTM, merge whenever you want. BTW if you want to cut a new release we'll merge it quickly in Firefox as there's a lot of desirable changes that have been made lately.

Jake-Shadle:
I can cut a release after lunch.

gabrielesvelto:
Thanks a lot :pray:

PR #135, opened 2024-09-04
Try to read the build id from the mapping file if reading from the process memory fails.

### LINKED ISSUE #134: Fall back to trying the file when retrieving the build id for a binary
If the build id of a binary isn't available in memory (which seems to at least sometimes be the case on Android), we should try to load the associated file to get the build id (as opposed to calculating one).

afranchuk:
We could potentially generalize this behavior to occur for SONAME as well, however it's likely unnecessary.

Closes #134.

### DIFF
--- a/src/linux/sections/mappings.rs
+++ b/src/linux/sections/mappings.rs
@@ -1,6 +1,6 @@
 use super::*;
 use crate::linux::maps_reader::MappingInfo;
-use crate::linux::module_reader::{BuildId, SoName};
+use crate::linux::module_reader::{BuildId, ReadFromModule, SoName};
 
 /// Write information about the mappings in effect. Because we are using the
 /// minidump format, the information about the mappings is pretty limited.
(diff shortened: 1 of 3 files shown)


src/linux/sections/mappings.rs
```diff
@@ -24,15 +24,41 @@ pub fn write(
...
+        log::debug!("retrieving build id for {:?}", &dumper.mappings[map_idx]);
         let BuildId(identifier) = dumper
             .from_process_memory_for_index(map_idx)
-            .unwrap_or_else(|_| BuildId(Vec::new()));
+            .or_else(|e| {
+                // If the mapping has an associated name that is a file, try to read the build id
+                // from the file. If there is no note segment with the build id in
+                // the program headers, we can't get to the note section if the section header
+                // table isn't loaded.
+                //
```
gabrielesvelto, line 35: nit: Remove this empty line

gabrielesvelto (approved):
LGTM, I just left a nit

PR #136, opened 2024-09-16
Use NonZeroUsize rather than generic NonZero, which is a fairly recent addition.

afranchuk:


Jake-Shadle:
Not sure I see the point of this? `NonZeroUsize` is literally just a type alias for `NonZero<usize>`.

afranchuk:
Generic `NonZero<T>` is only available since 1.79.0.

Jake-Shadle (approved):

PR #138, opened 2024-10-11
WIP: Add soft errors to minidump

### LINKED ISSUE #31: Add soft errors to minidump

marti4d:
Fixes #31

### DIFF
--- a/src/linux/minidump_writer.rs
+++ b/src/linux/minidump_writer.rs
@@ -2,11 +2,12 @@ pub use crate::linux::auxv::{AuxvType, DirectAuxvDumpInfo};
 use crate::{
     auxv::AuxvDumpInfo,
     dir_section::{DirSection, DumpBuf},
+    error_list::SoftErrorList,
     linux::{
         app_memory::AppMemoryList,
         crash_context::CrashContext,
         dso_debug,
-        errors::{InitError, WriterError},
+        errors::WriterError,
         maps_reader::{MappingInfo, MappingList},
         ptrace_dumper::PtraceDumper,
         sections::*,
@@ -144,8 +145,27 @@ impl MinidumpWriter {
             .clone()
             .map(AuxvDumpInfo::from)
             .unwrap_or_default();
-        let mut dumper = PtraceDumper::new(self.process_id, self.stop_timeout, auxv)?;
-        dumper.suspend_threads()?;
+
+        let mut soft_errors = SoftErrorList::default();
+
+        let (mut dumper, init_soft_errors) =
+            PtraceDumper::new_report_soft_errors(self.process_id, self.stop_timeout, auxv)?;
+
+        if !init_soft_errors.is_empty() {
+            soft_errors.push(WriterError::InitErrors(init_soft_errors));
+        }
+
+        let threads_count = dumper.threads.len();
+
+        if let Some(suspend_soft_errors) = dumper.suspend_threads().some() {
+            if dumper.threads.is_empty() {
+                // TBH I'm not sure this even needs to be a hard error. Is a minidump without any
+                // thread info still at-least a little useful?
+                return Err(WriterError::SuspendNoThreadsLeft(threads_count));
+            }
+            soft_errors.push(WriterError::SuspendThreadsErrors(suspend_soft_errors));
+        }
+
         dumper.late_init()?;
 
         if self.skip_stacks_if_mapping_unreferenced {
@@ -154,16 +174,15 @@ impl MinidumpWriter {
             }
 
             if !self.crash_thread_references_principal_mapping(&dumper) {
-                return Err(InitError::PrincipalMappingNotReferenced.into());
+                soft_errors.push(WriterError::PrincipalMappingNotReferenced);
             }
         }
 
         let mut buffer = Buffer::with_capacity(0);
-        self.generate_dump(&mut buffer, &mut dumper, destination)?;
+        self.generate_dump(&mut buffer, &mut dumper, soft_errors, destination)?;
 
-        // dumper would resume threads in drop() automatically,
-        // but in case there is an error, we want to catch it
-        dumper.resume_threads()?;
+        // TODO - Record these errors? Or maybe we don't care?
+        let _resume_soft_errors = dumper.resume_threads();
 
         Ok(buffer.into())
     }
@@ -226,11 +245,12 @@ impl MinidumpWriter {
         &mut self,
         buffer: &mut DumpBuf,
         dumper: &mut PtraceDumper,
+        mut soft_errors: SoftErrorList<WriterError>,
         destination: &mut (impl Write + Seek),
     ) -> Result<()> {
         // A minidump file contains a number of tagged streams. This is the number
         // of streams which we write.
-        let num_writers = 17u32;
+        let num_writers = 19u32;
 
         let mut header_section = MemoryWriter::<MDRawHeader>::alloc(buffer)?;
 
@@ -270,8 +290,12 @@ impl MinidumpWriter {
         let dirent = exception_stream::write(self, buffer)?;
         dir_section.write_to_file(buffer, Some(dirent))?;
 
-        let dirent = systeminfo_stream::write(buffer)?;
+        let (dirent, systeminfo_soft_errors) = systeminfo_stream::write(buffer)?;
         dir_section.write_to_file(buffer, Some(dirent))?;
+
+        if !systeminfo_soft_errors.is_empty() {
+            soft_errors.push(WriterError::WriteSystemInfoErrors(systeminfo_soft_errors));
+        }
 
         let dirent = memory_info_list_stream::write(self, buffer)?;
         dir_section.write_to_file(buffer, Some(dirent))?;
@@ -281,7 +305,10 @@ impl MinidumpWriter {
                 stream_type: MDStreamType::LinuxCpuInfo as u32,
                 location,
             },
-            Err(_) => Default::default(),
+            Err(e) => {
+                soft_errors.push(WriterError::WriteCpuInfoFailed(e));
+                Default::default()
+            }
         };
         dir_section.write_to_file(buffer, Some(dirent))?;
 
@@ -291,7 +318,10 @@ impl MinidumpWriter {
                 stream_type: MDStreamType::LinuxProcStatus as u32,
                 location,
             },
-            Err(_) => Default::default(),
+            Err(e) => {
+                soft_errors.push(WriterError::WriteThreadProcStatusFailed(e));
+                Default::default()
+            }
         };
         dir_section.write_to_file(buffer, Some(dirent))?;
 
@@ -303,7 +333,10 @@ impl MinidumpWriter {
                 stream_type: MDStreamType::LinuxLsbRelease as u32,
                 location,
             },
-            Err(_) => Default::default(),
+            Err(e) => {
+                soft_errors.push(WriterError::WriteOsReleaseInfoFailed(e));
+                Default::default()
+            }
         };
         dir_section.write_to_file(buffer, Some(dirent))?;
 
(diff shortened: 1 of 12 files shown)


marti4d:
Obviously the ([changes to minidump-common](https://github.com/rust-minidump/rust-minidump/pull/1035)) need to land before this can work.

This is WIP for now, but review comments would still be appreciated. I tried to at least turn the 3 things that @gabrielesvelto pointed out into soft errors, but I also included some other stuff that I thought could be reasonably recovered from.

I didn't want to get too carried away, but I think there's probably even more stuff we could turn into soft errors... But I think this provides a good starting point!

Curious to hear what you think :)

src/linux/minidump_writer.rs
```diff
@@ -144,8 +145,27 @@ impl MinidumpWriter {
...
+        if !init_soft_errors.is_empty() {
+            soft_errors.push(WriterError::InitErrors(init_soft_errors));
+        }
+
+        let threads_count = dumper.threads.len();
+
+        if let Some(suspend_soft_errors) = dumper.suspend_threads().some() {
+            if dumper.threads.is_empty() {
+                // TBH I'm not sure this even needs to be a hard error. Is a minidump without any
+                // thread info still at-least a little useful?
```
gabrielesvelto, line 163: Yes, it is still useful, but I think our tooling might not handle it well.

src/linux/minidump_writer.rs
```diff
@@ -383,3 +450,13 @@ impl MinidumpWriter {
...
+
+fn write_soft_errors(
+    buffer: &mut DumpBuf,
+    soft_errors: SoftErrorList<WriterError>,
+) -> Result<MDLocationDescriptor> {
+    let soft_error_list_str = format!("{soft_errors:?}");
+
+    let section = MemoryArrayWriter::write_bytes(buffer, soft_error_list_str.as_bytes());
+    Ok(section.location())
+}
```
gabrielesvelto, line 462: I like that this is another stream

src/linux/ptrace_dumper.rs
```diff
@@ -101,28 +152,56 @@ impl PtraceDumper {
...
+            }
+            _ => (),
+        }
+
+        // Same with mappings -- Some information is still better than no information!
+        match self.enumerate_mappings() {
+            Ok(()) => (),
+            Err(e) => {
+                soft_errors.push(InitError::EnumerateMappingsFailed(Box::new(e)));
+            }
```
gabrielesvelto, line 197: Looking how many times this pattern is repeated I wonder if it wouldn't be best to put the soft error list into the dumper. Then we could have a method that takes a result and records the error before returning, so we could write more straight-line code like:
```rust
let res: Result<Something, Error> = something_that_might_fail();
let foo = self.record_error(res)?;
// Do stuff with foo
```

marti4d: Totally fair -- I think you will find that it's a lot more concise now :)

src/linux/minidump_writer.rs
marti4d: 😁😁😁 - Don't forget that I need to get that added to `minidump-common` before I can merge this!

gabrielesvelto: Not necessarily, rust-minidump will ignore unknown streams, so we can land this if we want to, it's just that it won't be usable until we add support to `minidump-common`

gabrielesvelto: Oh wait, the definition of the stream is in `minidump-common`, silly me.

src/linux/ptrace_dumper.rs
```diff
@@ -101,23 +187,43 @@ impl PtraceDumper {
...
-        if let Err(e) = self.auxv.try_filling_missing_info(self.pid) {
-            log::warn!("failed trying to fill in missing auxv info: {e}");
+        // Even if we completely fail to fill in any additional Auxv info, we can still press
+        // forward.
+        if let Err(e) = self.auxv.try_filling_missing_info(
+            self.pid,
+            soft_errors.subwriter(InitError::FillMissingAuxvInfoErrors),
+        ) {
+            soft_errors.push(InitError::FillMissingAuxvInfoFailed(e));
+        }
```
gabrielesvelto, line 212: Just an idea here. Maybe it would be worth adding a function to  the list of errors so that this could be written as:
```rust
let res = self.auxv.try_filling_missing_info(self.pid, soft_errors.subwriter(InitError::FillMissingAuxvInfoErrors));
soft_errors.record_if_err(res); // or maybe_record_err(), whatever feels right
```

gabrielesvelto (changes requested):
This looks very nice! I've just left a few comments.

src/linux/minidump_writer.rs
```diff
@@ -154,16 +175,15 @@ impl MinidumpWriter {
...
 
         let mut buffer = Buffer::with_capacity(0);
-        self.generate_dump(&mut buffer, &mut dumper, destination)?;
+        self.generate_dump(&mut buffer, &mut dumper, soft_errors, destination)?;
 
-        // dumper would resume threads in drop() automatically,
-        // but in case there is an error, we want to catch it
-        dumper.resume_threads()?;
+        // TODO - Record these errors? Or maybe we don't care?
+        dumper.resume_threads(error_graph::strategy::DontCare);
```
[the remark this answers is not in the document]
marti4d: Ah sorry, it's not very obvious why this comment is in here, but the reason this is non-trivial is because the minidump has already been written by this point. I guess I would have to move this into `generate_dump()` after access to the thread information is no longer needed. I can do it, I just wanted to check with you that I should first?

src/linux/minidump_writer.rs
```diff
@@ -144,8 +147,26 @@ impl MinidumpWriter {
...
+
+        let threads_count = dumper.threads.len();
+
+        dumper.suspend_threads(soft_errors.subwriter(WriterError::SuspendThreadsErrors));
+
+        if dumper.threads.is_empty() {
+            // TBH I'm not sure this even needs to be a hard error. Is a minidump without any
+            // thread info still at-least a little useful?
+            return Err(WriterError::SuspendNoThreadsLeft(threads_count));
+        }
```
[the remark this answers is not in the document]
marti4d: Alrighty -- So leave it as-it and I'll just erase this question comment?

src/linux/ptrace_dumper.rs
marti4d: Oooo! I always like suggestions to improve ergonomics!

Your example is a good start,  but I think it only works as-written in this specific case, where the return type of the function is `Result<(), E>`. If we want it to be more generic to functions that return `Result<T, E>`, the API will need to be a bit more complex to pass forward the `Ok(T)` value.

For example, in our API above if `try_filling_missing_info()` actually returned a value (let's say `Result<Foo, AuxvError>`), we'd want our API to: map `Ok(Foo)` to `Foo`, map `Err(AuxvError)` to `AuxvError`, then map it with a user function to `InitError`, then push that onto the list, and finally return a `Foo` type.

So, I guess to do that, we'd have to have our new API take 2 functions: A function that maps `AuxvError` to `InitError` (which we already have, `InitError::FillMissingAuxvInfoFailed`) and a function that takes no arguments and produces a `Foo` (similar to `Option::unwrap_or_else()`). Normally the latter would just be `Foo::default` (We could also have a function similar to `unwrap_or_default()` if we want).

So, our new function (added to `trait WriteErrorList<E>`) would look something like this:

```Rust
fn unwrap_or_record_and_then<T, SubErr, SubMapFn, AndThenFn>(
    &mut self,
    result: Result<T, SubErr>,
    map_fn: SubMapFn,
    and_then: AndThenFn,
) -> T
where
    SubMapFn: FnOnce(SubErr) -> E,
    AndThenFn: FnOnce() -> T,
{
    match result {
        Ok(t) => t,
        Err(e) => {
            let parent_error = map_fn(e);
            self.push(parent_error);
            and_then()
        }
    }
}
```

And then here's what it would look like when it's used:

```Rust
let res = self.auxv.try_filling_missing_info(
    self.pid,
    soft_errors.subwriter(InitError::FillMissingAuxvInfoErrors),
);
let foo = soft_errors.unwrap_or_record_and_then(
    res,
    InitError::FillMissingAuxvInfoFailed,
    Foo::default);
```

Which just becomes your code example above in the special case that `T = ()`.

What do you think? Does it seem worth adding?

marti4d: Hmm... There is another way that might be even-more ergonomic -- Implement a trait directly on the `Result<T, E>` that allows recording it into a soft error. Then the code would look like this:

```Rust
let foo = self.auxv.try_filling_missing_info(
    self.pid,
    soft_errors.subwriter(InitError::FillMissingAuxvInfoErrors),
)
.unwrap_or_record_and_default(&mut soft_errors, InitError::FillMissingAuxvInfoFailed);
```

What do you think? Should we do the first one, this one, or neither?

src/linux/minidump_writer.rs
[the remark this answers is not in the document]
gabrielesvelto: Ah right, I hadn't thought about that. I was still thinking about the old scheme where we returned an error to the caller. Don't bother then, it's not very important.

src/linux/minidump_writer.rs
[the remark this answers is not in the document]
marti4d: Or did you mean that I should turn it into a soft error and continue to try and send the minidump?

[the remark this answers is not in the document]
marti4d: (I guess we'd still have the soft errors too, so... Maybe we should continue?)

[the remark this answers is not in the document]
marti4d: I just assumed that you wanted me to turn it into a soft error in my commit below.

src/linux/minidump_writer.rs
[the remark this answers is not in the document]
marti4d: Ah dang it -- I already did it :laughing:

src/linux/minidump_writer.rs
```diff
@@ -226,11 +246,12 @@ impl MinidumpWriter {
         &mut self,
         buffer: &mut DumpBuf,
         dumper: &mut PtraceDumper,
+        mut soft_errors: ErrorList<WriterError>,
         destination: &mut (impl Write + Seek),
     ) -> Result<()> {
         // A minidump file contains a number of tagged streams. This is the number
         // of streams which we write.
-        let num_writers = 17u32;
+        let num_writers = 19u32;
```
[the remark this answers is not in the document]
marti4d: Dammit! Good catch! As it turns out, "17 + 1 = 18"!

(This is my stupidity. At one point, I had 2 extra streams in the dump because I was writing both the JSON and the Debug-formatted version to the same file before I switched over to just the JSON.)

marti4d:
@gabrielesvelto -- I think that should address you feedback. I changed the count to "18" (whoops), made "no threads left" a soft error, and recorded soft errors from trying to restart the threads.

The only thing I'm not sure about is your ergonomics improvement suggestion. Personally, I think I like the idea of using a trait to add a new function to `Result<T, E>`, but I'm curious to hear if you think that it's worthwhile to do this?

src/linux/ptrace_dumper.rs
gabrielesvelto: `unwrap_or_record_and_default()` makes this very readable IMHO. Let's keep it for a subsequent iteration though, since you'd need to update the other crates too.

gabrielesvelto (approved):

marti4d:
@gabrielesvelto @Jake-Shadle This should be good to merge!

Jake-Shadle (approved):

PR #143, opened 2025-01-08
Update minidump crate versions

marti4d:


marti4d:
Thanks for that, Jake -- Once you land this, we should be good-to-go to rebase and land #138 as well. Then we can bask in soft-error glory! 😁

marti4d:
Also, #138 is failing the "deny-check" job as well. Is there some requirement you have at Embark that stops you from being able to use Cargo.lock version 4?

Jake-Shadle:
> Also, https://github.com/rust-minidump/minidump-writer/pull/138 is failing the "deny-check" job as well. Is there some requirement you have at Embark that stops you from being able to use Cargo.lock version 4?

The action is just outdated, I updated it in this PR.

PR #151, opened 2025-05-07
Use the debugger rendez-vous to populate the module list on Linux

### LINKED ISSUE #142: Use the debugger rendez-vous to populate the module list on Linux

marti4d:
Fixes #142

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ thiserror = "2.0"
 libc = "0.2"
 goblin = "0.9.2"
 memmap2 = "0.9"
+plain = { version = "0.2.3", default-features = false }
 
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
 nix = { version = "0.29", default-features = false, features = [
--- a/src/linux/maps_reader.rs
+++ b/src/linux/maps_reader.rs
@@ -111,25 +111,50 @@ pub enum MapsReaderError {
     SymlinkError(std::path::PathBuf, std::path::PathBuf),
 }
 
+/// Shared object loading information for the debugger
+///
+/// The Linux dynamic linker fills this info in the Dynamic section of the ELF headers at
+/// runtime. It is known as the "debugger rendez-vous" point and is a legacy structure to assist
+/// debuggers in locating loaded shared modules.
+///
+/// (But we're going to use it for minidump generation purposes.)
+///
+/// https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/link.h;h=b645760402514c4839686aaeade20dd5bb7725dd;hb=HEAD#l40
 #[allow(non_camel_case_types)]
 #[repr(C)]
 #[derive(Debug)]
 struct r_debug {
+    /// Version number of the protocol
     r_version: std::ffi::c_int,
+    /// Address of the first link_map structure
     r_map: usize,
+    /// Address of callback function for module map/unmap
     r_brk: usize,
+    /// Argument to r_brk on whether mapping is being added, subtracted, or is done
     r_state: u8,
+    /// Base address the linker is loaded at
     r_ldbase: usize,
 }
 
+/// Information for a single dynamically loaded module
+///
+/// This structure contains important information about a dynamically-loaded module, like its
+/// name and virtual address. It is also a node in a doubly-linked list of such modules.
+///
+/// https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/link.h;h=b645760402514c4839686aaeade20dd5bb7725dd;hb=HEAD#l101
 #[allow(non_camel_case_types)]
 #[repr(C)]
 #[derive(Debug)]
 struct link_map {
+    /// Base address module was loaded at
     l_addr: usize,
+    /// Name of file for module
     l_name: usize,
+    /// Address of the Dynamic section
     l_ld: usize,
+    /// Address of next node in list
     l_next: usize,
+    /// Address of previous node in list
     l_prev: usize,
 }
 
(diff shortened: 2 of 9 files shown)


marti4d:
@gabrielesvelto @Jake-Shadle You're going to want to review only [ea7b747](https://github.com/rust-minidump/minidump-writer/pull/151/commits/ea7b74755824b90d4d7ba09865031e6111353879). The only two commits are from the other two open PRs.

tests/linux_minidump_writer_failspot.rs
```diff
@@ -0,0 +1,40 @@
+#![cfg(any(target_os = "linux", target_os = "android"))]
```
marti4d, line 1: This test had to be put in its own file because current `failspot` tests cannot be run concurrently with non-`failspot` tests. I am hoping to change that eventually, but right now that's how it is...

gabrielesvelto:
Gonna review this ASAP. Thank you @marti4d

Cargo.toml
```diff
@@ -28,6 +28,7 @@ thiserror = "2.0"
 libc = "0.2"
 goblin = "0.9.2"
 memmap2 = "0.9"
+plain = { version = "0.2.3", default-features = false }
```
gabrielesvelto, line 31: This is one neat crate! I'll probably make further use of it.

src/linux/mem_reader.rs
```diff
@@ -105,26 +109,84 @@ impl MemReader {
...
+            v.push(self.read_pod(address)?);
+            address += std::mem::size_of::<T>();
+        }
+        Ok(v)
+    }
+
+    pub fn read_until(
+        &mut self,
+        mut address: usize,
+        byte: u8,
```
gabrielesvelto, line 143: nit: Let's call this argument `terminator` to make the intent clearer

src/linux/ptrace_dumper.rs
```diff
@@ -433,23 +443,14 @@ impl PtraceDumper {
...
-        self.mappings = MappingInfo::aggregate(maps, self.auxv.get_linux_gate_address())
-            .map_err(InitError::AggregateMappingsFailed)?;
+    fn enumerate_mappings(
+        &mut self,
+        mut soft_errors: impl WriteErrorList<InitError>,
+    ) -> Result<(), InitError> {
+        let mut mappings = self.enumerate_mappings_via_proc_maps().or_else(|e| {
+            soft_errors.push(e);
+            self.enumerate_mappings_via_debugger_rendezvous()
+        })?;
```
gabrielesvelto, line 453: IIUC we're trying to enumerate the mappings using the debugger rendez-vous only if reading them from the proc map fails, right? I'd like this to be more granular if possible, because the debugger rendez-vous should have more reliable information with regards to loaded module, but as we discussed, it should also be missing things like JIT areas.

Would it be possible to first read the module list using the debugger rendez-vous, then augment it with the entries found in `/proc/<pid>/maps` that do not correspond to the already existing module entries? If we fail to read modules via the debugger rendez-vous, or if it's missing some, those will be picked up from `/proc/<pid>/maps` instead.

The reason I'd like it done this way is that the merging logic in `MappingInfo::aggregate()` is finicky and caused us problems in the past, and the information present in-memory should be the ultimate source of truth regarding how far a module extends in memory (and what it refers to).

src/linux/ptrace_dumper.rs
```diff
@@ -464,16 +465,77 @@ impl PtraceDumper {
...
+
+        let Some(program_header_count) = self
+            .auxv
+            .get_program_header_count()
+            .map(|x| usize::try_from(x).unwrap())
+        else {
+            return Err(InitError::MissingProgramHeaderCount);
+        };
+
+        Self::suspend_thread(self.pid).map_err(InitError::SuspendMainThreadFailed)?;
```
gabrielesvelto, line 500: I think this needs a bit of high-level refactoring. IIRC we start a dump by attempting to stop the process in `PtraceDumper::init()` which we call when we create the object. Then we call this still from within `PtraceDumper::init()` and then we resume that thread later, but then we stop all the threads again when the code in `MinidumpWriter::dump()` calls `PtraceDumper::suspend_threads()` directly and finally we resume all the threads at the end of `MinidumpWriter::generated_dump()` and finally resume the process. Given that we'd like for the process to be stopped for the entire lifetime of the `PtraceDumper` object let's do it this way (unless there's a downside I'm not aware of):
* When start constructing the `PtraceDumper` object by stopping the process and then suspending all threads (and we start recording soft failures for those)
* All the methods in `PtraceDumper` can then assume that all threads are stopped (we know sometimes they're not, but still). This should make things like enumerating stuff under `/proc` safer too, as we can't race with the threads before suspension.
* Finally, let's make `PtraceDumper::generate_dump()` consume the object, so when we resume all the threads at the end the user can't call anything on it anymore (and the destructor will still do the resume threads / resume the process dance).

WDYT?

gabrielesvelto (changes requested):

src/linux/mem_reader.rs
marti4d: I changed it. I had actually mostly named it this to be inline with the [`std` trait function of the same name](https://doc.rust-lang.org/stable/std/io/trait.BufRead.html#method.read_until)

src/linux/maps_reader.rs
```diff
@@ -130,6 +205,187 @@ impl MappingInfo {
...
+            validate_elf_signature(&module_elf_header)?;
+
+            let module_program_header_virtual_address =
+                usize::try_from(module_elf_header.e_phoff).unwrap();
+            let module_program_header_address =
+                module_elf_header_address + module_program_header_virtual_address;
+            let module_program_header_len = usize::from(module_elf_header.e_phnum);
+            let module_program_headers: Vec<ProgramHeader> = memory_reader
+                .read_pod_vec(module_program_header_address, module_program_header_len)
+                .map_err(E::ReadProgramHeaderTableFailed)?;
```
gabrielesvelto, line 319: As with in the previous function you can break this out into a smaller function to make the logic simpler, e.g. `let module_program_headers = read_module_program_headers(...)?`.

src/linux/maps_reader.rs
```diff
@@ -130,6 +205,187 @@ impl MappingInfo {
...
+            for module_program_header in module_program_headers
+                .iter()
+                .filter(|x| x.p_type == PT_LOAD)
+            {
+                let segment_virtual_address =
+                    usize::try_from(module_program_header.p_vaddr).unwrap();
+                let segment_address = module_elf_header_address + segment_virtual_address;
+                let segment_len = usize::try_from(module_program_header.p_memsz).unwrap();
+                let segment_end_address = segment_address + segment_len;
+                let segment_offset = usize::try_from(module_program_header.p_offset).unwrap();
```
gabrielesvelto, line 330: `let segment_offset = calculate_segment_offset(...)`

src/linux/maps_reader.rs
```diff
@@ -130,6 +205,187 @@ impl MappingInfo {
...
+
+                // Round each of these down or up to the nearest page
+                let segment_address = segment_address / 4096 * 4096;
+                let segment_offset = segment_offset / 4096 * 4096;
+                let segment_end_address = segment_end_address.div_ceil(4096) * 4096;
+                let segment_len = segment_end_address - segment_address;
```
gabrielesvelto, line 336: Is the page size guaranteed to always be 4KiB? It'd be better to use a constant for it and explain why it can't change.

src/linux/maps_reader.rs
```diff
@@ -130,6 +205,187 @@ impl MappingInfo {
...
+                let mut permissions = MMPermissions::empty();
+                if module_program_header.p_flags & PF_R != 0 {
+                    permissions.insert(MMPermissions::READ);
+                }
+                if module_program_header.p_flags & PF_W != 0 {
+                    permissions.insert(MMPermissions::WRITE);
+                }
+                if module_program_header.p_flags & PF_X != 0 {
+                    permissions.insert(MMPermissions::EXECUTE);
+                }
```
gabrielesvelto, line 347: `let permissions = from_program_header(...)`

gabrielesvelto (changes requested):
Mostly stylistic/readability changes but I was also curious about the hard-coded page size in the code.

gabrielesvelto (approved):
Looking good!

gabrielesvelto:
@marti4d do you want to double-check the Android failures or shall I merge?

marti4d:
I am going to spend some time trying to look into them, but it's weird. I don't see the issue in my local emulator, or on real Android devices. So it makes me wonder if there's something going on with CI, or if the emulator is just running too slow on weak hardware?

Either way, maybe I should try even more to figure it out before we merge.

src/linux/maps_reader.rs
```diff
@@ -98,6 +111,93 @@ pub enum MapsReaderError {
...
+#[derive(Debug)]
+struct r_debug {
+    /// Version number of the protocol
+    r_version: std::ffi::c_int,
+    /// Address of the first link_map structure
+    r_map: usize,
+    /// Address of callback function for module map/unmap
+    r_brk: usize,
+    /// Argument to r_brk on whether mapping is being added, subtracted, or is done
+    r_state: u8,
```
schopin-mozilla, line 134: Looking at the bionic source code, this field is an i32, not a u8:

https://android.googlesource.com/platform/bionic/+/a27d2baa/linker/linker.h#49

gabrielesvelto: Oh my, it's an enum in glibc so... I guess it's different between Android and Linux? Since it only has three value I guess that GCC defaults this as a u8 when compiling glibc but I'd like to verify that first.

gabrielesvelto: I played a little bit with godbolt and both clang and GCC will lay out `r_state` as a 32-bit field on Linux too, so let's go for that.

schopin-mozilla: I managed to reproduce the hang locally, so I tried changing that field to a `i32` and the hang disappeared. Now I'm chasing a new failure :)

marti4d:
Superseded by #195

src/linux/maps_reader.rs
```diff
@@ -107,12 +109,93 @@ pub enum MapsReaderError {
...
+/// runtime. It is known as the "debugger rendez-vous" point and is a legacy structure to assist
+/// debuggers in locating loaded shared modules.
+///
+/// (But we're going to use it for minidump generation purposes.)
+///
+/// https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/link.h;h=b645760402514c4839686aaeade20dd5bb7725dd;hb=HEAD#l40
+#[allow(non_camel_case_types)]
+#[repr(C)]
+#[derive(Debug)]
+struct r_debug {
```
schopin-mozilla, line 126: @marti4d I know it's been a while since you wrote this, but is there a reason you didn't use the similar definitions in `src/linux/dso_debug.rs`? I just stumbled upon them by accident.

marti4d: Honestly? I was very new to `minidump-writer` at that point and probably didn't realize it was even there. I probably did a quick grep for `r_debug`, didn't see any hits, and wrote my own.

But yep -- Now that I know it's there, we should probably just prefer to use that instead.

PR #155, opened 2025-07-11
Merge PtraceDumper and MinidumpWriter

### LINKED ISSUE #154: Merge `PtraceDumper` and `MinidumpWriter`
When this project was first created, it probably made a lot of sense for `PtraceDumper` and `MinidumpWriter` to be two different types in case there was ever a "not-ptrace dumper".

At this point, however, the functioning of these two structs is so closely coupled that it's doubtful that real cut lines could actually be made between them without major refactoring work anyway. In the meantime, any time we work on the code there's a weird moment of "Umm... Which struct should this thing be a part of?", and the answer is often "whichever struct already has code that kind-of looks like it".

I ran into that pretty hard when working on #142, where I got annoyed by trying to break my code change up along the seemingly-artificial line between the two. I finally decided to do a table flip and did a bunch of refactoring work. I hope y'all like it!

marti4d:
Merges PtraceDumper and MinidumpWriter

While I'm at it, also reorganize a bunch of code and update some things to use emerging Rust best-practices.

- Change use statements to use blocks
- Prefer to use `super::` rather than `crate::` where applicable
- Use mod.rs in a directory instead of <modname>.rs when a module has submodules
- Move error type definitions closer to where they are used
  - Most error types are pretty specific to a tight scope, so put them there
  - Exception is the MinidumpWriter's error type - There's so much code that it's better to put it in its own module
- Merge PtraceDumper and MinidumpWriter
  - The split between them has been blurry for a while
  - Merge them and their error types together
  - A nice thing about the split was that `MinidumpWriter` had a bit of a "builder config" behavior for `PtraceDumper`. Let's keep that!
  - There is now a `MinidumpWriterConfig` that is explicitly a fluent API used to configure `MinidumperWriter`
- Change functions that were tightly coupled to `MinidumpWriter` to just be impls on it
- Update tests to accept new structure

Fixes #154


src/linux/minidump_writer/mod.rs
```diff
@@ -1,212 +1,268 @@
...
-    let pid = nix::unistd::Pid::from_raw(child);
-    ptrace::detach(pid, None).or_else(|e| {
-        // errno is set to ESRCH if the pid no longer exists, but we don't want to error in that
-        // case.
-        if e == nix::Error::ESRCH {
-            Ok(())
-        } else {
-            Err(DumperError::PtraceDetachError(child, e))
+impl MinidumpWriterConfig {
+    pub fn new(process_id: Pid, blamed_thread: Pid) -> Self {
```
gabrielesvelto, line 124: Just a thought for the future: we do this in pretty much all our crates but we should really have a separate `Tid` types for threads. Obviously something for a follow-up.

gabrielesvelto:
Thanks @Jake-Shadle! I'm merging it

PR #157, opened 2025-09-09
Use shared references for MemReader methods

marti4d:
[`OnceLock`](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html) is perfect for situations like this, where a one-time memoization needs to be done and then the object will never mutate again.

Now we can use shared references for `MemReader`, which is nice because sometimes it can be awkward not being able to read two areas of the other process's memory concurrently (despite there being no reason it can't be done).

### DIFF
--- a/src/linux/mem_reader.rs
+++ b/src/linux/mem_reader.rs
@@ -1,7 +1,11 @@
 //! Functionality for reading a remote process's memory
 
-use super::{minidump_writer::MinidumpWriter, serializers::*, Pid};
+use {
+    super::{minidump_writer::MinidumpWriter, serializers::*, Pid},
+    std::sync::OnceLock,
+};
 
+#[derive(Debug)]
 enum Style {
     /// Uses [`process_vm_readv`](https://linux.die.net/man/2/process_vm_readv)
     /// to read the memory.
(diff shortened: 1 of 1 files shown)


src/linux/mem_reader.rs
```diff
@@ -1,7 +1,11 @@
 //! Functionality for reading a remote process's memory
 
-use super::{minidump_writer::MinidumpWriter, serializers::*, Pid};
+use {
+    super::{minidump_writer::MinidumpWriter, serializers::*, Pid},
+    std::sync::OnceLock,
+};
 
+#[derive(Debug)]
```
gabrielesvelto, line 8: Is the `Debug` trait required or can it be removed? I wouldn't add it unless the structure is user-printable.

gabrielesvelto (approved):
Excellent, just one nit

src/linux/mem_reader.rs
marti4d: Yep -- This is needed to be able to use `.expect()` on the potential `Err(Style)` that might be returned by `OnceLock::set()`. Alternatively, I could use `.map(|_| ()).expect(...)` on it to avoid this, but honestly this is an internal type anyway - What's the point of avoiding `#[derive(Debug)]` on it?

PR #164, opened 2025-12-16
Avoid attempting to read `/etc/(?:os | lsb)-release` on Android

### LINKED ISSUE #163: Don't try to read the Linux distribution information on Android
As per title. Every Android minidump will have a soft-error logged because it couldn't find the `lsb-release` or `os-release` files, but that's expected so we should not even try to read them there (see the code [here](https://github.com/rust-minidump/minidump-writer/blob/ad678474221bf4cd8e658337d0c6bbef181101c5/src/linux/minidump_writer/mod.rs#L396-L408)).

Jake-Shadle:
Resolves: #163

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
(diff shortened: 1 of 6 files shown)


Jake-Shadle:
I noticed the Android tests were failing non-deterministically, in different tests, so assumed it was some kind of poor timing/concurrency issue in the emulator, changing to running Android tests sequentially with `-j 1` seems to make it behave.

gabrielesvelto:
> I noticed the Android tests were failing non-deterministically, in different tests, so assumed it was some kind of poor timing/concurrency issue in the emulator, changing to running Android tests sequentially with `-j 1` seems to make it behave.

Oh thank you! That's something that blocked https://github.com/rust-minidump/minidump-writer/pull/151 from landing, because we didn't know if it was the PR's fault or something pre-existing.

.github/workflows/ci.yml
gabrielesvelto, line 1: Nice!

src/windows/minidump_writer.rs
```diff
@@ -119,14 +119,13 @@ impl MinidumpWriter {
             };
 
             let mut exception_record: EXCEPTION_RECORD = std::mem::zeroed();
+            exception_record.ExceptionCode = exception_code;
```
gabrielesvelto: This could be written:
```rust
let mut exception_record = EXCEPTION_RECORD {
  ExceptionCode: exception_code,
 ..std::mem::zeroed()
}
```

gabrielesvelto (changes requested):
Just one nit but the rest is fine, thanks for all the cleanups

Jake-Shadle:
Damn, looks like I didn't actually fix the Android tests. I think it's the linux gate and mapping test that often fail, and it feels like it's not our fault, so maybe we just disable those tests for android only?

gabrielesvelto:
> Damn, looks like I didn't actually fix the Android tests. I think it's the linux gate and mapping test that often fail, and it feels like it's not our fault, so maybe we just disable those tests for android only?

Yes, let's file an issue to investigate it later. It's not a big deal anyway.

gabrielesvelto (approved):

PR #167, opened 2025-12-25
Ensure all ptrace() and /proc/<pid> IO is through an interface

### LINKED ISSUE #166: Minidump-Writer won't work in an Android Isolated Process
See [Firefox bug 1710940 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1710940) for full details, but basically if the process that crashes is an Android Isolated Process, it has its own ephemeral UID and SELinux context that prevents external processes from using `ptrace()` on it or reading its `/proc/<pid>` files unless the external process is also in that same UID and SELinux context.

That's all a very fancy way of saying that a crashed process can basically only be debugged by itself or a child process.

Since Firefox runs minidump-writer in a separate privileged process, minidump-writer is unable to dump the crashed process.

To fix this, we can add infrastructure to `minidump-writer` to allow it to send requests for `ptrace()` and IO to `/proc/<pid>` to a small "dummy" child process that is forked from the crashed process and get the child to act on its behalf.

marti4d:
When Android Isolated Process is enabled, every process exists in its own ephemeral UID and SELinux context. That will essentially block `ptrace()` and I/O to `/proc/<pid>`, which means minidump-writer currently can't be used to dump Isolated Processes.

This change refactors existing code to do all forbidden operations through a trait, `ProcessInspector`, that currently only has one implementation - `DirectInspector` - which calls the APIs directly (like the current implementation does).

In the next change, we will add another implementation of `ProcessInspector` - `remote::InspectorInvoker` and its other half `remote::InspectorExecutor` - That will allow minidump-writer to send `ptrace()` and `/proc/<pid>` I/O requests to a remote process that does have the required permissions.

Fixes #166


src/linux/thread_info/mod.rs
```diff
@@ -81,44 +75,11 @@ pub fn copy_u32_registers(dst: &mut [u128], src: &[u32]) {
...
-                format!("/proc/{tid}/status"),
-                ppid,
-                tgid,
-            ));
-        }
-        Ok((ppid, tgid))
-    }
+// TODO - Does this ptrace(PTRACE_GETREGS) and ptrace(PTRACE_GETFPREGS) stuff need to be
+// abstracted too? Breakpad doesn't do it, but it seems like the type of thing that
+// should be stopped by Isolated Processes...
```
gabrielesvelto, line 80: Yes, given the isolation is around the uid/gid then these calls should also fail unless they're proxy-ed.

src/linux/dso_debug.rs
```diff
@@ -97,8 +98,8 @@ pub struct RDebug {
 }
 
 pub fn write_dso_debug_stream(
+    process_inspector: &mut dyn ProcessInspector,
```
gabrielesvelto, line 101: I noticed you're declaring this argument as mutable here and in several other places, but from what I can tell it doesn't need to be mutable. Surprisingly clippy doesn't complain :shrug:. Is this because of the proxy-ed reader?

gabrielesvelto (commented):
This looks like a refactoring of the existing code but without introducing the new proxy-based `ProcessInspector`. These changes look fine the way they are but did you forget adding the new implementation to the PR?

afranchuk:
@marti4d is there more work to do here? I'd like to get it in so that #168 can be applied on top.

PR #168, opened 2025-12-30
Bug 1831094: Add a process reader and module reader for each platform

afranchuk:
This merges features from the firefox process_reader crate for [bug 1831094](https://bugzilla.mozilla.org/show_bug.cgi?id=1831094), in a step toward using `minidump-writer` to read crash annotations on its own.

### DIFF
--- a/src/mac/module_reader.rs
+++ b/src/mac/module_reader.rs
@@ -1,0 +1,173 @@
+use crate::module_reader::{ModuleMemory, ModuleMemoryReadError};
+use crate::serializers::*;
+use goblin::container::{Container, Ctx, Endian};
+use goblin::mach::{
+    self,
+    header::{Header, Header64, MH_DYLIB, MH_EXECUTE, MH_MAGIC_64},
+    load_command::{LC_SEGMENT_64, LoadCommandHeader, Section64, SegmentCommand64},
+};
+use scroll::ctx::{SizeWith, TryFromCtx};
+
+const DATA_SEGMENT: &[u8; 16] = b"__DATA\0\0\0\0\0\0\0\0\0\0";
+
+pub struct ModuleReader<'a> {
+    module_memory: ModuleMemory<'a>,
+    header: Header,
+    context: Ctx,
+}
+
+#[derive(Debug, thiserror::Error, serde::Serialize)]
+pub enum ModuleReaderError {
+    #[error(transparent)]
+    ReadModuleMemory(#[from] ModuleMemoryReadError),
+    #[error("failed to parse MachO memory: {0}")]
+    GoblinParsing(
+        #[from]
+        #[serde(serialize_with = "serialize_generic_error")]
+        goblin::error::Error,
+    ),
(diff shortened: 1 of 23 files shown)


gabrielesvelto:
Sorry I've been holding off reviewing this one but I was waiting for PR https://github.com/rust-minidump/minidump-writer/pull/167 first given the changes there are more urgent and both PRs end up touching the same functionality.

afranchuk:
Sounds good, I'll rebase once that is merged.

marti4d:
@afranchuk I was talking to @gabrielesvelto yesterday. I don't want to hold up changes to `minidump-writer` any longer on my change. I keep thinking that I'm close to done only for there to be more undiscovered complexity. And I'm also going to be on PTO soon for a few weeks... I just don't think it makes sense for you to wait any longer.

Honestly -- If I have to redo the part where I go through the existing codebase and redirect all the I/O, that's not the major difficulty of the change anyway -- Everything else is 😆.

src/module_reader.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl<'a> ModuleMemory<'a> {
+    pub fn from_process(reader: &'a ProcessReader, start_address: usize) -> Self {
+        Self::Process {
+            reader,
+            start_address: start_address as u64,
+        }
+    }
+
+    #[inline]
```
gabrielesvelto, line 65: I know these were in the original code but are the `#[inline]` directives really useful?

src/mac/module_reader.rs
```diff
@@ -0,0 +1,173 @@
...
+        let mut vals = Vec::with_capacity($count);
+        let mut offset = 0;
+        for _ in 0..$count {
+            let (value, bytes_read) = <$T>::try_from_ctx(&bytes[offset..], $ctx)?;
+            offset += bytes_read;
+            vals.push(value);
+        }
+        vals
+    }};
+}
```
gabrielesvelto, line 59: Is there any particular reason why you used `scroll` specifically for the macOS module reader but not the other platforms?

gabrielesvelto (approved):
LGTM, I just left a couple of questions

src/module_reader.rs
afranchuk: For this method at least, perhaps not. Maybe it's to encourage the compiler to elide the enum variant checks to have less branching on that? I can see an argument for wanting reading bytes to be optimized. But then again, there's plenty of error-checking that needs to be done here anyway, so there's still overhead for every call.

src/mac/module_reader.rs
afranchuk: It wasn't very intentional: we already had code for linux, so I didn't touch that much, and on windows the `goblin` interface provides some convenience methods to get what we need. The `goblin` macOS interface wasn't as convenient (it doesn't have a lazy interface), so I read it directly instead.

afranchuk:
Good to land this?

gabrielesvelto:
Yes, this is good to go.

PR #169, opened 2026-01-09
Update to 2024 edition

Jake-Shadle:
The 2024 edition has been out for almost a year, so this just updates to it and sets the MSRV to 1.85.0, which is the version that eg. Debian packages.


Cargo.lock
```diff
@@ -9,7 +9,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "fallible-iterator 0.3.0",
- "gimli",
+ "gimli 0.31.1",
```
gabrielesvelto, line 12: Ouch, lots of duplications, not just this one. I'll try to set aside some time to send PRs upstream to the offending crates because it will make it harder for us to vendor it.

gabrielesvelto (approved):
~~LGTM, thanks!~~

gabrielesvelto (changes requested):
Ugh, I hit the approve button too fast and I hadn't looked at the automation. This still needs changes on macOS and Windows to build.

Cargo.lock
Jake-Shadle: Yes, I don't like the duplicates, but I think the usual usage of this crate is fine since all/most of them are only due to dev-dependencies.

src/bin/test.rs
```diff
@@ -340,14 +340,33 @@ mod linux {
...
+                .expect("failed to read /dev/urandom");
+
+            for index in rand_indices {
+                name.push(CHARS[index as usize % CHARS.len()] as char);
+            }
+
+            write!(&mut name, "{id}").unwrap();
+
+            let path = root.join(name);
+            file_array.push(std::fs::File::create(&path).expect("failed to create path"));
```
gabrielesvelto, line 373: This is breaking the tests on Android. I guess it needs a separate code path. If it's too much work just use `tempfile` again.

Jake-Shadle: I think this new commit fixes it, I removed `tempfile` since it's really a dev-dependency, but due to how the test binary works it needed to be a regular dependency, but it introduced several duplicates, notably `windows-sys`.

gabrielesvelto (approved):
Thanks!

PR #172, opened 2026-05-05
Updated several dependencies

gabrielesvelto:
- This fixes all the issues reported by `cargo audit`
- Several duplicate dependencies have been removed
- Minimum rust version is bumped to 1.88.0 because of the `time` crate

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1384,7 +1384,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "uuid",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Jake-Shadle:
I wasn't really paying attention to #168, and this PR was partially broken due to updating the minor version of windows-sys, we already have inline bindings to the functionality that we needed before that PR in https://github.com/rust-minidump/minidump-writer/blob/main/src/windows/ffi.rs so that this crate isn't a contributor to the windows-sys version issue, would you be against me adding the few structs/functions to that and getting rid of the windows-sys dependency? (in a separate PR)

Jake-Shadle (approved):

gabrielesvelto:
That would be fine! The less windows-sys we have the better

PR #173, opened 2026-05-05
Remove direct dependency on windows-sys

Jake-Shadle:
This replaces the windows-sys dependency added in #168 with inline bindings so that this crate won't be a contributor to duplicate versions of windows-sys.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1384,7 +1384,6 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "uuid",
- "windows-sys 0.52.0",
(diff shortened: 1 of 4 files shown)


gabrielesvelto (approved):
LGTM, thanks!

gabrielesvelto:
@Jake-Shadle could you please cut a new release now that this has been merged? I'd like to update our vendored crate and there's a lot of work coming down the pipe that will change the Linux dumper to make it work on Android isolated process. That's going to be fairly intrusive and I'd rather keep that in a future release.

PR #174, opened 2026-05-06
Isolated Processes 1: Refactor thread_info for ease of redirecting

marti4d:
Currently, thread_info is a bit of a mess and is difficult to cleanly redirect into an interface.

This cleans it up and makes it easier to move around.

It also fixes a piece of unsound code (mod.rs line 165) where we are passing an immutable pointer to an API that says it will modify the result.

### DIFF
--- a/src/linux/thread_info/aarch64.rs
+++ b/src/linux/thread_info/aarch64.rs
@@ -1,7 +1,6 @@
 use {
-    super::{CommonThreadInfo, NT_Elf, Pid, ThreadInfoError},
+    super::{Pid, PtraceRequestType, ThreadInfoError},
     crate::minidump_cpu::{FP_REG_COUNT, GP_REG_COUNT, RawContextCPU},
-    nix::sys::ptrace,
 };
 
 /// https://github.com/rust-lang/libc/pull/2719
@@ -25,47 +24,29 @@ pub struct ThreadInfoAarch64 {
     pub fpregs: user_fpsimd_struct,
 }
 
-impl CommonThreadInfo for ThreadInfoAarch64 {}
-
 impl ThreadInfoAarch64 {
     pub fn get_instruction_pointer(&self) -> usize {
         self.regs.pc as usize
     }
 
-    // nix currently doesn't support PTRACE_GETREGSET, so we have to do it ourselves
     fn getregset(pid: Pid) -> Result<libc::user_regs_struct> {
-        Self::ptrace_get_data_via_io(
-            0x4204 as ptrace::RequestType, // PTRACE_GETREGSET
-            Some(NT_Elf::NT_PRSTATUS),
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const NT_PRSTATUS: usize = 1;
+        super::ptrace_getregset(NT_PRSTATUS, pid)
     }
 
     fn getregs(pid: Pid) -> Result<libc::user_regs_struct> {
-        // TODO: nix restricts PTRACE_GETREGS to arm android for some reason
-        Self::ptrace_get_data(
-            12 as ptrace::RequestType, // PTRACE_GETREGS
-            None,
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const PTRACE_GETREGS: PtraceRequestType = 12;
+        unsafe { super::ptrace_getregs::<libc::user_regs_struct>(PTRACE_GETREGS, pid) }
     }
 
-    // nix currently doesn't support PTRACE_GETREGSET, so we have to do it ourselves
     fn getfpregset(pid: Pid) -> Result<user_fpsimd_struct> {
-        Self::ptrace_get_data_via_io(
-            0x4204 as ptrace::RequestType, // PTRACE_GETREGSET
-            Some(NT_Elf::NT_PRFPREGSET),
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const NT_PRFPREGSET: usize = 2;
+        super::ptrace_getregset(NT_PRFPREGSET, pid)
     }
 
-    // nix currently doesn't support PTRACE_GETFPREGS, so we have to do it ourselves
     fn getfpregs(pid: Pid) -> Result<user_fpsimd_struct> {
-        Self::ptrace_get_data(
-            14 as ptrace::RequestType, // PTRACE_GETFPREGS
-            None,
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const PTRACE_GETFPREGS: PtraceRequestType = 14;
+        unsafe { super::ptrace_getregs::<user_fpsimd_struct>(PTRACE_GETFPREGS, pid) }
     }
 
     pub fn fill_cpu_context(&self, out: &mut RawContextCPU) {
@@ -85,8 +66,8 @@ impl ThreadInfoAarch64 {
         out.float_regs[..FP_REG_COUNT].copy_from_slice(&self.fpregs.vregs[..FP_REG_COUNT]);
     }
 
-    pub fn create_impl(_pid: Pid, tid: Pid) -> Result<Self> {
-        let (ppid, tgid) = Self::get_ppid_and_tgid(tid)?;
+    pub fn create(_pid: Pid, tid: Pid) -> Result<Self> {
+        let (ppid, tgid) = super::get_ppid_and_tgid(tid)?;
         let regs = Self::getregset(tid).or_else(|_| Self::getregs(tid))?;
         let fpregs = Self::getfpregset(tid).or_else(|_| Self::getfpregs(tid))?;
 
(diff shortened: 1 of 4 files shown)


src/linux/thread_info/mod.rs
```diff
@@ -48,147 +52,82 @@ cfg_if::cfg_if! {
...
-        let io = libc::iovec {
-            iov_base: data.as_mut_ptr().cast(),
-            iov_len: std::mem::size_of::<T>(),
-        };
-        let res = unsafe {
-            libc::ptrace(
-                request,
-                libc::pid_t::from(pid),
-                flag.unwrap_or(NT_Elf::NT_NONE),
-                &io as *const _,
```
marti4d, line 162: Note that this is likely UB, as `io` is declared immutable, this is a `*const T`, but `ptrace()` says it will update `iov_len` with the actual number of bytes read.

Also -- We should check the actual number of bytes read and make sure it's `size_of::<T>()`

src/linux/thread_info/mod.rs
```diff
@@ -48,147 +52,82 @@ cfg_if::cfg_if! {
...
-        }
-        Ok((ppid, tgid))
     }
-
-    /// SLIGHTLY MODIFIED COPY FROM CRATE nix
-    /// Function for ptrace requests that return values from the data field.
-    /// Some ptrace get requests populate structs or larger elements than `c_long`
-    /// and therefore use the data field to return values. This function handles these
-    /// requests.
-    fn ptrace_get_data<T>(
```
marti4d, line 124: Also note that this function should have always been marked as `unsafe` -- If the size of the value passed to `request` doesn't match with the size of `T`, it is UB due to uninitialized memory if T is too big or UB due to buffer overflow if T is too small.

src/linux/thread_info/aarch64.rs
```diff
@@ -25,47 +24,29 @@ pub struct ThreadInfoAarch64 {
...
     }
 
-    // nix currently doesn't support PTRACE_GETREGSET, so we have to do it ourselves
     fn getregset(pid: Pid) -> Result<libc::user_regs_struct> {
-        Self::ptrace_get_data_via_io(
-            0x4204 as ptrace::RequestType, // PTRACE_GETREGSET
-            Some(NT_Elf::NT_PRSTATUS),
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const NT_PRSTATUS: usize = 1;
```
gabrielesvelto, line 33: `goblin` already defines this so we might as well use that definition instead of rolling our own

src/linux/thread_info/arm.rs
```diff
@@ -30,25 +29,15 @@ pub struct ThreadInfoArm {
...
     }
 
-    // nix currently doesn't support PTRACE_GETREGS, so we have to do it ourselves
     fn getregs(pid: Pid) -> Result<user_regs_struct> {
-        Self::ptrace_get_data::<user_regs_struct>(
-            ptrace::Request::PTRACE_GETREGS as ptrace::RequestType,
-            None,
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const PTRACE_GETREGS: PtraceRequestType = 12;
```
gabrielesvelto, line 39: This is defined in the version of the `libc` crate we vendor with Firefox so we can use that definition instead.

src/linux/thread_info/mips.rs
```diff
@@ -1,54 +0,0 @@
...
-            out.float_save.regs[idx] = self.mcontext.fpregs.fp_r.fp_fregs[idx]._fp_fregs;
-        }
-
-        out.float_save.fpcsr = mcontext.fpc_csr;
-
-        // #if _MIPS_SIM == _ABIO32
-        //   out.float_save.fir = self.mcontext.fpc_eir;
-        // #endif
-    }
-}
```
gabrielesvelto: IIUC the MIPS code originates from the very first version of minidump-writer and was ported over from Breakpad. I doubt anybody ever used it so I'm OK with removing it given we have no way to test it. @Jake-Shadle is this OK with you?

That being said if we remove the MIPS code I'd like to see it done in a separate patch that comes before this one.

src/linux/thread_info/mod.rs
```diff
@@ -48,147 +52,82 @@ cfg_if::cfg_if! {
...
+                    .ok_or_else(|| ThreadInfoError::InvalidProcStatusFile(tid, l.clone()))?
+                    .parse::<Pid>()?;
             }
+            "PPid:\t" => {
+                ppid = l
+                    .get(6..)
+                    .ok_or_else(|| ThreadInfoError::InvalidProcStatusFile(tid, l.clone()))?
+                    .parse::<Pid>()?;
+            }
+            _ => continue,
```
gabrielesvelto, line 88: You could use the functionality offered by the `procfs-core` crate for parsing this but it's not a blocker. Maybe let's make a note to move this in a separate issue.

src/linux/thread_info/x86.rs
```diff
@@ -101,107 +107,59 @@ pub struct ThreadInfoX86 {
...
-
 impl ThreadInfoX86 {
-    // nix currently doesn't support PTRACE_GETREGSET, so we have to do it ourselves
     fn getregset(pid: Pid) -> Result<user_regs_struct> {
-        Self::ptrace_get_data_via_io(
-            0x4204 as ptrace::RequestType, // PTRACE_GETREGSET
-            Some(NT_Elf::NT_PRSTATUS),
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const NT_PRSTATUS: usize = 1;
```
gabrielesvelto, line 117: Same comment as above WRT not redefining this.

src/linux/thread_info/x86.rs
```diff
@@ -101,107 +107,59 @@ pub struct ThreadInfoX86 {
...
     }
 
     pub fn getregs(pid: Pid) -> Result<user_regs_struct> {
-        // TODO: nix restricts PTRACE_GETREGS to arm android for some reason
-        Self::ptrace_get_data(
-            12 as ptrace::RequestType, // PTRACE_GETREGS
-            None,
-            nix::unistd::Pid::from_raw(pid),
-        )
+        const PTRACE_GETREGS: PtraceRequestType = 12;
```
gabrielesvelto, line 122: Likewise, I think this should be always defined in libc now.

src/linux/thread_info/x86.rs
```diff
@@ -394,3 +352,30 @@ impl ThreadInfoX86 {
...
+    }
+}
+
+#[cfg(target_arch = "x86_64")]
+pub fn copy_u32_registers(dst: &mut [u128], src: &[u32]) {
+    // SAFETY: We are copying a block of memory from ptrace as u32s to the u128
+    // format of minidump-common
+    unsafe {
+        let dst: &mut [u8] =
+            std::slice::from_raw_parts_mut(dst.as_mut_ptr().cast(), dst.len() * 16);
```
gabrielesvelto: use `size_of::<u128>()` instead of 16

src/linux/thread_info/x86.rs
```diff
@@ -394,3 +352,30 @@ impl ThreadInfoX86 {
...
+        let src: &[u8] = std::slice::from_raw_parts(src.as_ptr().cast(), src.len() * 4);
```
gabrielesvelto: use `size_of::<u32>()` instead of 4

gabrielesvelto (changes requested):
Small changes here and there, most of this LGTM

src/linux/thread_info/mips.rs
marti4d: Good idea -- Broke it out into PR #178 and merged it already.

src/linux/thread_info/mod.rs
marti4d: I didn't write this code, just moved it around, and rewriting this seems a bit more risky than some of the other refactoring I did, so yeah... I'd rather leave it for a separate change.

src/linux/thread_info/aarch64.rs
marti4d: So I started doing this, but surprisingly `libc` and `goblin` have some of these constants and not others. So I did this to favour consistency instead of having a bit of a hodge podge of "use the `libc`/`goblin` constant if it exists, otherwise just define it ourselves".

If you think that's the wrong tradeoff though, that's fair -- I'll make the change. I just like this better so all the functions read the same.

src/linux/thread_info/mod.rs
marti4d: Filed issue #179

src/linux/thread_info/x86.rs
marti4d: I decided to go with the surprise third option: rewrite this algorithm to be way nicer to read.

gabrielesvelto (approved):
LGTM, if @Jake-Shadle is OK with it too I'd merge it.

marti4d:
Thanks @gabrielesvelto! @Jake-Shadle was originally a reviewer here and removed himself, so I'm going to assume that was him deferring to your judgement for the review -- So I'll merge this in :)

PR #175, opened 2026-05-07
Isolated Processes 2: Redirect thread_info I/O and ptrace

marti4d:
thread_info is complex enough to redirect that it could use its own PR for review.

### DIFF
--- a/src/linux/minidump_writer/mod.rs
+++ b/src/linux/minidump_writer/mod.rs
@@ -7,6 +7,7 @@ use {
         dso_debug,
         dumper_cpu_info::CpuInfoError,
         maps_reader::{MappingInfo, MappingList, MapsReaderError},
+        process_inspection::ProcessInspector,
         process_reader::{CopyFromProcessError, ProcessReader},
         serializers::*,
         thread_info::{ThreadInfo, ThreadInfoError},
@@ -41,8 +42,6 @@ use {
 
 #[cfg(target_os = "android")]
 use super::android::late_process_mappings;
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-use super::thread_info;
 
 pub use super::auxv::{AuxvType, DirectAuxvDumpInfo};
 
@@ -82,6 +81,7 @@ pub struct MinidumpWriterConfig {
     crashing_thread_context: CrashingThreadContext,
     stop_timeout: Duration,
     direct_auxv_dump_info: Option<DirectAuxvDumpInfo>,
+    process_inspector: ProcessInspector,
 }
 
 #[derive(Debug)]
@@ -104,6 +104,7 @@ pub struct MinidumpWriter {
     pub crash_context: Option<CrashContext>,
     pub app_memory: AppMemoryList,
     pub memory_blocks: Vec<MDMemoryDescriptor>,
+    process_inspector: ProcessInspector,
 }
 
 #[derive(Debug, Clone)]
@@ -137,6 +138,7 @@ impl MinidumpWriterConfig {
             crashing_thread_context: Default::default(),
             stop_timeout: STOP_TIMEOUT,
             direct_auxv_dump_info: Default::default(),
+            process_inspector: ProcessInspector::local(),
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/linux/minidump_writer/mod.rs
```diff
@@ -515,7 +518,11 @@ impl MinidumpWriter {
     }
 
     /// Suspends a thread by attaching to it.
-    fn suspend_thread(child: Pid) -> Result<(), WriterError> {
+    fn suspend_thread(
+        #[allow(unused)] // TODO
```
gabrielesvelto: Add a reason such as
```rust
#[allow(unused, "Only required by x86 to inspect stack pointers")]
```
or something along the lines

src/linux/thread_info/x86.rs
```diff
@@ -113,41 +17,25 @@ pub struct ThreadInfoX86 {
...
-        unsafe { super::ptrace_getregs::<user_fpxregs_struct>(PTRACE_GETFPXREGS, pid) }
-    }
-
-    pub fn create(_pid: Pid, tid: Pid) -> Result<Self> {
-        let (ppid, tgid) = super::get_ppid_and_tgid(tid)?;
-        let regs = Self::getregset(tid).or_else(|_| Self::getregs(tid))?;
-        let fpregs = Self::getfpregset(tid).or_else(|_| Self::getfpregs(tid))?;
+    pub fn create(
+        process_inspector: &ProcessInspector,
+        _pid: Pid,
```
gabrielesvelto: This parameter appears unused both here and in the other implementations, I'd say we should remove it entirely.

src/linux/process_inspection/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+fn getregset(_pid: libc::pid_t) -> nix::Result<GenRegs> {
+    #[cfg(target_arch = "arm")]
+    {
+        Err(Errno::ENOTSUP)
+    }
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
+    {
+        const NT_PRSTATUS: usize = 1;
```
gabrielesvelto, line 74: Like I mentioned in the previous patch IIRC this is already defined in `goblin` so we could use that.

src/linux/process_inspection/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+        ptrace_getregset(NT_PRSTATUS, _pid)
+    }
+}
+
+fn getregs(pid: libc::pid_t) -> nix::Result<GenRegs> {
+    const PTRACE_GETREGS: PtraceRequestType = 12;
```
gabrielesvelto, line 80: This should be defined on every platform in `libc` IIRC.

gabrielesvelto (changes requested):
LGTM with some tiny changes.

src/linux/minidump_writer/mod.rs
marti4d: This is... Extremely temporary. In the next patch, it will get used... a lot... on all platforms.

marti4d: I guess I could just silence the warning by starting the variable with an underscore, so I just did that instead. That way, it will naturally fix itself when I actually use it in the next change.

src/linux/thread_info/x86.rs
marti4d: Good catch! Got rid of it.

src/linux/process_inspection/mod.rs
marti4d: See my previous comment I guess about the "consistency <-> library-reuse" tradeoff that's being made here.

gabrielesvelto (approved):

marti4d:
@Jake-Shadle I don't want to keep assuming -- Do you want to also review these isolated process changes, or are you fine with just @gabrielesvelto doing them?

Jake-Shadle:
Oh sorry, I'm completely out of the loop on these, so don't wait on my review.

PR #178, opened 2026-05-08
Remove references to unsupported architectures

marti4d:
There are references to PowerPC and MIPS in the code, but why bother mentioning all the architectures we don't support?

MIPS technically has some code in it, but it's not being tested and doesn't even compile, and you know the old saying: "If it hasn't been tested, it doesn't work."


Jake-Shadle (approved):

marti4d:
Thanks Jake! I'll just merge this in then :)

PR #183, opened 2026-05-15
Enable support for FreeBSD

### LINKED ISSUE #162: Please add support for FreeBSD
Coming from https://github.com/EmbarkStudios/crash-handling/issues/108#issuecomment-3497203941

> FreeBSD support must be added first to the [minidump-writer](https://github.com/rust-minidump/minidump-writer/) crate. There's nothing at the moment but part of the reason why we started that crate, using it to replace Breakpad, is that we wanted to make it easier to add minidump generation for BSDs. I haven't looked into how much work that's going to be, but it should be given that FreeBSD has ptrace() it might not be too much work. FYI I don't have any spare cycles for it, but I'm very happy to mentor, review and merge contributions if someone has them.

The FreeBSD port of [zed editor](https://zed.dev/) is stuck at v0.198.6 due to this issue:

> Committer's Note: The port is at version 0.198.6, whereas the latest version is 0.212.5. Later versions than 0.198.6 depend on "crash-handler" crate, which does not currently support FreeBSD.
-- https://www.freshports.org/editors/zed

G36maid:
CLOSES #162

WIP

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,15 @@ jobs:
     strategy:
       matrix:
         job:
-          - { os: ubuntu-24.04,   target: x86_64-unknown-linux-gnu, release: true }
-          - { os: ubuntu-24.04,   target: x86_64-unknown-linux-musl }
-          - { os: windows-2025,   target: x86_64-pc-windows-msvc }
+          - {
+              os: ubuntu-24.04,
+              target: x86_64-unknown-linux-gnu,
+              release: true,
+            }
+          - { os: ubuntu-24.04, target: x86_64-unknown-linux-musl }
+          - { os: windows-2025, target: x86_64-pc-windows-msvc }
           - { os: macos-15-intel, target: x86_64-apple-darwin }
-          - { os: macos-15,       target: aarch64-apple-darwin }
+          - { os: macos-15, target: aarch64-apple-darwin }
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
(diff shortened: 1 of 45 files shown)


gabrielesvelto:
Thank you for this work, FreeBSD support in this crate together with some upstream work I'm making in Firefox would make it possible to have proper crash reporting enabled on the platform. It's worth mentioning that we're refactoring the interfaces we use to read from a process to make room from a proxy'd process reader which we need on Android. Check the changes in both the pending and already merged PRs that are called "Isolated Processes X: ...". Those changes are largely mechanical and do not affect the logic, just where some pieces live and how they're accessed.

G36maid:
Thanks for the quick response! It's great to know this work could help.

I'm currently running some bare-metal testing on FreeBSD — things are looking good so far. I'd be happy if you'd like to start reviewing this PR while I finish up the testing.

I've also taken a peek at the Isolated Processes PR series (#/174–#/182). Once the testing is done and this PR is in a good state, I plan to follow up by adapting ProcessInspector for FreeBSD to align with the new architecture.

 One question: would you prefer I do it as a follow-up PR after this one lands, or fold it in here? I 'd prefer to keep them separate — this PR is already fairly large. Happy to go either way though.

gabrielesvelto:
> One question: would you prefer I do it as a follow-up PR after this one lands, or fold it in here? I 'd prefer to keep them separate — this PR is already fairly large. Happy to go either way though.

It's fine to make that a follow up if it makes life easier for you.

G36maid:
Sorry for the delay over the past few weeks. I was caught up with some unexpected interviews and tests, so I didn't have time to push this PR forward. 

I have some free time now, so I'll review where I left off and pick this back up.

gabrielesvelto:
np, I've been very busy myself and I also need to spin up a FreeBSD VM to test

gabrielesvelto:
Quick update: I still haven't found the time to spin up a FreeBSD VM because I've been constantly putting out fires at work and churning through other reviews. I should have some time the next week to look into this, or at the absolute worst at the end of the month. I wish I could be more responsive to contributions because they genuinely interest me and I'm always happy to receive them but work has been absolutely hectic leaving me with very little time and energy for anything else.

G36maid:
No worries at all! That actually works out because I haven't finished organizing and polishing up my side of the PR yet either.

PR #184, opened 2026-06-01
Isolated Processes 6: Add a basic libc backend

marti4d:


Cargo.toml
```diff
@@ -1,14 +1,33 @@
...
+
+[workspace.package]
 repository = "https://github.com/rust-minidump/minidump-writer"
 homepage = "https://github.com/rust-minidump/minidump-writer"
 edition = "2024"
 rust-version = "1.88.0"
 license = "MIT"
 
+[workspace.dependencies]
+libc = "0.2"
```
gabrielesvelto, line 15: Shouldn't this be gated by `target.'cfg(unix)'` since we don't use it on Windows?

crates/linux/process-backend/src/local/syscall_runner.rs
```diff
@@ -0,0 +1,46 @@
+use {super::errno, core::ffi::c_int};
+
+#[derive(Debug, Default)]
+pub struct SyscallRunner(Option<c_int>);
```
gabrielesvelto: I'm a bit on edge about this being called a "runner", it gives me the impression that it's doing something asynchronous which it is not. I understand that `SyscallCaller` sounds redundant, so what about `SyscallInvoker` and then have `invoke()` instead of `run()`?

gabrielesvelto (changes requested):

Cargo.toml
marti4d: So... I thought and expected that too! ... But, surprisingly, that's not how workspace stuff works.

Apparently the best way to think of these `workspace.dependencies` and `workspace.package` tables is just to think of them as nothing but an inert "Single-Source of Truth Key-Value Store" that does nothing by itself.

It's only when the values in the table are actually referenced by setting `field.workspace = true` that the value is actually substituted in, and it's done in the context where it's requested.

So if nothing on Windows actually calls for a dependency, it will be ignored. This is actually true of the `workspace.members` as well -- I have `crates/linux/process-backend` listed as a workspace member, but if nobody actually depends on it, it won't be compiled.

crates/linux/process-backend/src/local/syscall_runner.rs
marti4d: Good suggestion! Done!

gabrielesvelto (approved):

PR #185, opened 2026-06-01
Isolated Processes 7: Rewrite all backend functions in libc

marti4d:


crates/linux/process-backend/src/local/mod.rs
```diff
@@ -57,10 +344,115 @@ impl Backend {
     }
 }
 
+#[derive(Debug)]
+pub struct FileReader(OwnedFd);
+
+impl FileReader {
+    pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
+        let rv = unsafe { libc::read(self.0.0, buf.as_mut_ptr().cast(), buf.len()) };
```
gabrielesvelto: I wonder if it would be worth making `FileReader` `#[repr(transparent)]` to avoid the double `.0` field access?

crates/linux/process-backend/src/local/mod.rs
```diff
@@ -57,10 +344,115 @@ impl Backend {
...
+        if rv == -1 {
+            return Err(Error::ReadFileFailed(errno()));
+        }
+        Ok(rv.try_into().unwrap())
```
gabrielesvelto, line 356: Just a nit, for things that return `ssize_t` it might be worth changing the check into `rv < 0` and then you can do an `unwrap_unchecked()` when converting to `size_t`.

crates/linux/process-backend/src/local/module_reader.rs
```diff
@@ -0,0 +1,152 @@
...
+        // zero-length mappings -- So we just make it a one-byte mapping (and ignore the byte).
+        let len = usize::max(len, 1);
+
+        // Rust/LLVM cannot support a single object larger than `isize::MAX`, which is
+        // 2GiB on 32-bit systems. It is possible to map files larger than that, but there is a
+        // bunch of special handling that needs to be done to avoid accidentally telling LLVM that
+        // the mapped memory might be a single object with the same provenance.
+        //
+        // Luckily, we won't be accessing files that are larger than 2GiB, so we can skip all that
+        // nastiness by disallowing a mapping larger than `isize::MAX`.
```
gabrielesvelto, line 107: Do you have a reference for this to add here? I'm not familiar with this limitation.

src/linux/minidump_writer/mod.rs
```diff
@@ -948,21 +949,12 @@ impl MinidumpWriter {
     #[inline]
     pub fn copy_from_process(
         process_inspector: &ProcessInspector,
-        pid: Pid,
+        _pid: Pid,
```
gabrielesvelto: We can just remove this, right?

gabrielesvelto (approved):
LGTM, I just left a couple of nits here and there

crates/linux/process-backend/src/local/mod.rs
marti4d: Yeah 😅 I hate when I see double `.0` as well -- It was my fault for being a bit lazy with the way I implemented `OwnedFd` more than a need for this to be `Deref<Target = OwnedFd>` (which is what I think you wanted -- `#[repr(transparent)]` doesn't flatten the object; it just makes its binary representation identical).

I added a proper `as_raw_fd()` function to `OwnedFd` to more closely mimic what they do in `std`, and also added an `unsafe fn new()` to it so it properly shows that creating an `OwnedFd` is a dangerous operation.

crates/linux/process-backend/src/local/mod.rs
marti4d: Hmm... I'm not so sure about this suggestion. All these OS APIs claim that they return `-1` to indicate error, not "a negative number".

So, to me, this code says exactly that:

| RV | Meaning |
|---|---|
| >= 0 | Successful return value |
| -1 | Failure |
| Anything Else | Software bug - Let's panic! |

crates/linux/process-backend/src/local/module_reader.rs
marti4d: Good idea! Added links to the relevant `std` docs :)

src/linux/minidump_writer/mod.rs
marti4d: Ah -- Very good point! I thought we needed it to have the same shape as it does for Windows and Darwin (a lot of the Memory/Module reading stuff is like that), but it turns out... Nope. So I got rid of it.

Although - as you can see - the word "just" was understating it a bit 😆.

crates/linux/process-backend/src/local/mod.rs
gabrielesvelto: OK, fair enough. In another time and place I would have a assumed that such a syscall would always behave in a predictable way, but with people vibe-coding the kernel I don't know what to think anymore. Better safe than sorry.

gabrielesvelto (approved):

PR #186, opened 2026-06-02
Isolated Processes 8: Remove dependency on nix

marti4d:
The `nix` crate is barely used now that all this stuff has been rewritten in `libc` -- Let's deliver the final blow to it!

This actually makes life a lot easier on the Firefox side of things, because auditing new versions of the `nix` crate is awful -- The entire thing is practically nothing but `unsafe` code, and we have to audit the entire crate even if we only use a handful of functions from it because our approval affects the safety of everyone who uses our whitelist on `cargo vet`.

### DIFF
--- a/tests/linux_minidump_writer_soft_error.rs
+++ b/tests/linux_minidump_writer_soft_error.rs
@@ -58,7 +58,7 @@ fn soft_error_stream_content() {
                     }"
                 }
             ]},
-            {"SuspendThreadsErrors": [{"PtraceAttachError": [1234, "EPERM"]}]}
+            {"SuspendThreadsErrors": [{"PtraceAttachError": [1234, 1]}]}
         ]}),
         json!({"WriteSystemInfoErrors": [
             {"WriteCpuInformationFailed": {
--- a/tests/linux_minidump_writer_soft_error.rs
+++ b/tests/linux_minidump_writer_soft_error.rs
@@ -58,7 +58,7 @@ fn soft_error_stream_content() {
                     }"
                 }
             ]},
-            {"SuspendThreadsErrors": [{"PtraceAttachError": [1234, 1]}]}
+            {"SuspendThreadsErrors": [{"PtraceAttachError": [1234, libc::EPERM]}]}
         ]}),
         json!({"WriteSystemInfoErrors": [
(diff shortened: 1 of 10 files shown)


tests/linux_minidump_writer_soft_error.rs
```diff
@@ -58,7 +58,7 @@ fn soft_error_stream_content() {
                     }"
                 }
             ]},
-            {"SuspendThreadsErrors": [{"PtraceAttachError": [1234, "EPERM"]}]}
+            {"SuspendThreadsErrors": [{"PtraceAttachError": [1234, 1]}]}
```
gabrielesvelto: Why not `libc::EPERM`?

tests/ptrace_dumper.rs
```diff
@@ -188,8 +186,9 @@ fn linux_gate_mapping_id() {
 
 #[test]
 fn merges_mappings() {
-    let page_size = nix::unistd::sysconf(nix::unistd::SysconfVar::PAGE_SIZE).unwrap();
-    let page_size = std::num::NonZeroUsize::new(page_size.unwrap() as usize).unwrap();
+    let page_size = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+    assert!(page_size > 0);
+    let page_size = std::num::NonZeroUsize::new(page_size as usize).unwrap();
```
gabrielesvelto: The `unwrap()` is redundant given you've already checked that `page_size` is positive, maybe do `NonZeroUsize::new_unchecked()`?

gabrielesvelto (approved):
LGTM, left just a couple of nits.

tests/linux_minidump_writer_soft_error.rs
marti4d: Lol good catch -- That was silly of me 😅

tests/ptrace_dumper.rs
marti4d: Honestly? Now that I look, I don't really think that the whole `NonZeroUsize` thing is really pulling its weight in this function anymore. When `nix` needed it, sure... But when we're directly calling `libc` ourselves now... Nah. It's better in interfaces than it is in the middle of a function. I just changed it to use raw `usize` now.

PR #187, opened 2026-06-12
Restore the linux ProcessReader API

afranchuk:
This API was added in #168 and is present in 0.12, however it was lost in #181. It is necessary for [Bug 2009827 - Use minidump-writer in place of the in-tree process_reader crate](https://bugzilla.mozilla.org/show_bug.cgi?id=2009827).


afranchuk:
Note that it would be better to move the ProcessReader stuff into a separate crate, however until we have time to do that (which will take a little effort as there is definitely some tangling in other code), this exposes what we need.

afranchuk:
@marti4d I can't add you as a reviewer, but take a look.

gabrielesvelto:
Ah yes, this was my mistake, but there's an extra angle to this: not only we need the interface but we need it to work with isolated processes too. If that's not possible we need to fold the annotation-reading code into the regular dumping process (potentially as a feature since we'd be the only ones interested in it).

gabrielesvelto:
Adding Chris as a reviewer.

marti4d:
Superseded by #188

PR #188, opened 2026-06-24
Isolated Processes 9: Repair ProcessReader API

marti4d:


marti4d:
@afranchuk I also can't add you as a reviewer for whatever reason, but... Tada! 🎉

crates/process-reader/src/linux/error.rs
afranchuk, line 1: Any reason to avoid `thiserror` (it is `no_std`)? I don't mind them being written out, I'm just curious. I keep wishing they'd just include it in `core`, but that's not likely to happen.

crates/process-reader/src/linux/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ReadError`] if the selected strategy reports an error, or if
+    /// automatic strategy selection cannot get any strategy to return success.
+    ///
+    /// A failed read may have partially overwritten `buf`, and the error does not
+    /// report how many bytes were copied before the failure. Callers should not
+    /// rely on the contents of `buf` after an error.
+    pub fn read_at(&self, address: usize, buf: &mut [u8]) -> Result<usize, ReadError> {
```
afranchuk, line 342: Maybe add a convenience `read_exact_at` for the common case where you want to fill the buffer?

crates/process-reader/src/linux/wrapper.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+}
+
+impl Drop for File {
+    fn drop(&mut self) {
+        unsafe {
+            let rv = libc::close(self.0);
+            if rv == -1 {
+                // Not much we can really do here
```
afranchuk, line 85: You could add the `log` crate, but since this is the only such place that has a "dead-end", it's unnecessary. You could also add an explicit, fallible `File::close()` method, but again, probably unnecessary for our needs.

crates/process-reader/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "process-reader"
```
afranchuk, line 2: Since this crate is intended for only linux and android, I'd call this `process-reader-linux`.

Or: broaden the scope of the crate, move this out of the `linux` subdirectory, and leave macos/windows impls empty for the moment (those could be moved from `minidump_writer` later).

src/linux/process_inspection/mod.rs
```diff
@@ -21,53 +19,58 @@ pub mod process_reader;
 #[derive(Debug)]
 pub struct ProcessInspector {
     pid: libc::pid_t,
-    process_reader: ProcessReader,
-    backend: Rc<Backend>,
+    backend: Backend,
 }
 
 #[derive(Debug)]
 pub enum Backend {
```
afranchuk, line 26: For the methods that are expected to be the same on each backend, lifting them to `impl Backend` itself will consolidate the `match`es to once place instead of each call site (and it'll also give a minimum template for the requirements of each backend).

afranchuk (changes requested):
This looks great! Let's rename or broaden the `process-reader` crate.

crates/process-reader/src/linux/error.rs
marti4d: Not a big deal either way -- I know some people in the Rust community get a little bit like "What? You're going to make me use proc macros and increase my compile time by a bunch just so you can avoid writing a handful of error strings?"

... But yeah, I agree that `thiserror` should just being in the dang standard library lol. I dunno -- If you think I should do just use `thiserror` then I'll do it, but... I've already done the work.

crates/process-reader/src/linux/wrapper.rs
marti4d: FWIW I had the exact same thought, and I think I even did this in my `libc` wrappers elsewhere in the code.

Again it was really just about avoiding a dependency just for a one-liner.

crates/process-reader/Cargo.toml
marti4d: Ah -- Yeah, good point for sure. I think I will move it up a directory and we can eventually add other implementations to it :)

src/linux/process_inspection/mod.rs
marti4d: It's not up for review yet, but I've already destroyed all of this and replaced it with a trait object. But yeah, point taken that if I was going to keep it like that, I could at-least make it a little less "explosion of match statements everywhere" that way.

crates/process-reader/src/linux/error.rs
afranchuk: Nah, it's already all written out :shrug:

src/linux/process_inspection/mod.rs
afranchuk: Ah good, I was wondering where the trait object that you had mentioned before ended up!

crates/process-reader/Cargo.toml
marti4d: Done!

crates/process-reader/src/linux/mod.rs
marti4d: Yep yep -- Good idea. Done!

marti4d:
@gabrielesvelto This should be a pretty quick review, since Alex has already approved it :)

crates/process-reader/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+#![no_std]
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub use linux::*;
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+mod linux;
```
afranchuk: I usually do something like the following, which avoids the duplicated `cfg`s.
```
#[cfg_attr(any(target_os = "linux", target_os = "android"), path = "linux/mod.rs")]
mod platform;

pub use platform::*;
```

afranchuk (approved):

gabrielesvelto:
Looking at this now. I assume it obsoletes PR https://github.com/rust-minidump/minidump-writer/pull/187 right?

gabrielesvelto (approved):
LGTM, thanks to both of you

crates/process-reader/src/lib.rs
marti4d: Ah -- Good call! It's funny - I've seen this done in other code dozens of times and never internalized it, but yeah... This is way better. Done!

PR #190, opened 2026-07-15
Move crash-context to this repo

Jake-Shadle:
It makes much more sense for crash-context to live here since it's so closely tied to minidump-writer, in addition to the work of adding a new platform to it requiring work in minidump-writer as well before it is actually useful.


gabrielesvelto (approved):
LGTM. Thanks this will make our lives simpler!

PR #194, opened 2026-07-21
tests: allow testing on arm64 Android emulators

schopin-mozilla:
This is notably useful when using an Apple Silicon machine for development.

### DIFF
--- a/src/linux/dumper_cpu_info/arm.rs
+++ b/src/linux/dumper_cpu_info/arm.rs
@@ -1,6 +1,7 @@
 use {
     super::{CpuInfoError, ProcessInspector},
     crate::minidump_format::*,
+    failspot::failspot,
     scroll::Pwrite,
     std::{
         collections::HashSet,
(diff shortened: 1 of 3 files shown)


src/linux/dumper_cpu_info/arm.rs
```diff
@@ -191,9 +192,17 @@ pub fn write_cpu_information(
...
+    }
+
     let cpuinfo_file = match process_inspector.read_file("/proc/cpuinfo") {
         Ok(x) => x,
-        Err(_) => {
+        Err(e) => {
+            // Surface a failspot-injected failure as a soft error.
+            if failspot!(CpuInfoFileOpen) {
+                return Err(CpuInfoError::ReadFileError(e));
+            }
```
marti4d: Ah, this is interesting -- I did not notice before that we swallow the error here on ARM but not on x86.

I think this comment is stale, actually, and it's fine to just unconditionally return the error now. You can see in the [x86 version of this](https://github.com/rust-minidump/minidump-writer/blob/main/src/linux/dumper_cpu_info/x86.rs#L52) that we've just flattened it to a `?` operator.

And the reason we can do that is because [we always treat this function failing as a soft error](https://github.com/rust-minidump/minidump-writer/blob/main/src/linux/minidump_writer/systeminfo_stream.rs#L35). That wasn't always the case, but it is now.

marti4d (changes requested):

src/linux/dumper_cpu_info/arm.rs
schopin-mozilla: II'm guessing that I first wrote the patch against an old version of main, since it was to test #151. This should now be fixed, I aligned the code with the x86 version.

marti4d (approved):
LGTM! Okay if I merge this, @schopin-mozilla?

schopin-mozilla:
I'm hoping you will, since I can't 😉

Jake-Shadle (approved):

schopin-mozilla:
The formatting should be fixed now.

PR #195, opened 2026-07-22
Use the debugger rendez-vous to populate the module list on Linux

### LINKED ISSUE #142: Use the debugger rendez-vous to populate the module list on Linux

schopin-mozilla:
I'm taking #151 over from Chris.

Fixes #142.

### DIFF
--- a/.cargo/android-runner.sh
+++ b/.cargo/android-runner.sh
@@ -3,7 +3,11 @@
 BINARY=$1
 shift
 
+# Cargo doesn't expose the target triple to the runner, so derive it from the
+# binary path (.../target/<triple>/...).
+TARGET=$(printf '%s' "$BINARY" | sed -E 's#(.*/)?target/([^/]+)/.*#\2#')
+
 # Make sure to run the following to copy the test helper binary over.
-# cargo run --target x86_64-linux-android --bin test
+# cargo run --target "$TARGET" --bin test
 adb push "$BINARY" "/data/local/$BINARY"
-adb shell "chmod 777 /data/local/$BINARY && env TEST_HELPER=/data/local/target/x86_64-linux-android/debug/test /data/local/$BINARY" "$@"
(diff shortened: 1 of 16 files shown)


schopin-mozilla:
The branch is based on top of #194 to make my life easier debugging Android issues, hence the Draft status, although I'm not sure if the CI runs on draft PRs?

schopin-mozilla:
I think I'm getting there in terms of understanding this code, and I think we
might need to change the approach a bit.

AFAICT, the reason we *want* to use the rendezvous data is to generate a more
accurate ModulesList stream. What that means exactly remains to be determined: 

* Do we want the exact executable code mapping, single-byte precision, or are
  we aiming for a broader data set that includes alignment-induced padding? The
  current procmaps code seems to be in the latter case, judging by the aggregate()
  algorithm.

* do we want to include .bss sections and similar, or limit ourselves to
  executable stuff? In the current code, pretty much anything that's explicitly
  file-backed (which excludes .bss and similar sections) and contiguous to a
  code+padding section gets thrown in.

Another question: do we really want this as a *fallback*? Instead, I'd rather
have this data as primary source, but used *only* to populate the ModulesList.
As it stands, the data structure that is populated in the current implementation
is also used for other purposes, where that data source is a pretty poor fit:

* Thread stacktraces. In order to walk the stacks, we must first identify the stack
  mappings. The new data source doesn't know about those since it's not ELF-derivated.
* IP lookups. On its face it seems reasonable to expect the instruction pointer to
  point within an ELF codepage, but then you remember about JIT-generated code...

My proposed new implementation plan is thus the following:

- Keep the current procmap-based data source for general mapping data
- Use the rendezvous data source to populate a separate module data structure,
  with a fallback to the previous procmaps approach
- Add a config knob to force the use of the fallback, to make it easy for us
  to have a progressive rollout (credits to @marti4d for that idea)

Assuming you'll agree with that plan, I'll still need clarification on *which*
data we want in the modules list to finalize the work, but I can already get
started on the overall architecture.

schopin-mozilla:
Closing this in favor of #234 and friends.

PR #196, opened 2026-07-23
Fix test flakiness on Linux

schopin-mozilla:
Some of our tests build crash contexts for the *parent* process, to then write the minidump for a *child* process. Most of the time, the instruction and stack pointers will just point to unmapped memory, which is just silently ignored, but because of ASLR the memory layouts are random and it happens that sometimes they point to mapped but unreadable memory, leading to hard errors.

This branch addresses the issue on both ends: we build a zeroed out crash context (at least we know *how* it's invalid), and we also use soft errors rather than hard ones to be consistent with the unmapped experience.

FWIW, the unreadable case happens pretty frequently on Android ARM64 for some reason.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,6 +11,7 @@ cfg_if::cfg_if! {
                 CpuInfoFileOpen,
                 StackPointerMapping,
                 ThreadStackCopy,
+                CrashingThreadIpCopy,
             }
         }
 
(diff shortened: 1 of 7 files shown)


marti4d (approved):
Awesome! Glad to see tests got added for this stuff, and thanks for breaking it apart into those 4 commits -- Was much easier to review than the whole thing at once.

I'll merge this in now :)

PR #197, opened 2026-08-06
CI: ARM native tests and platform-specific Clippy

schopin-mozilla:
This PR adds Linux arm64 in the set of main targets (GH has had runners for a while now), and runs clippy for all those main targets rather than just for x86_64-unknown-linux-gnu, since we have a sizable amount of platform-specific code that deserves to be checked.

And of course, bundled in are the various clippy fixes for the new failures those changes brought forward.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,8 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt,clippy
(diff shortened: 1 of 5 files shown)


schopin-mozilla:
The GH action doesn't appear to pull the appropriate target. Marking it as Draft while I'm investigating.

gabrielesvelto (approved):
LGTM

PR #198, opened 2026-08-10
Isolated Processes 10: Refactor ProcessReader

marti4d:
The way ProcessReader was implemented (creating a new one when a test requests it) was never a great approach, but it gets even worse when we can't allocate memory in the executor process.

This refactors it to simply hold one ProcessReader at a time, which can be changed by a client that has an exclusive reference to the backend.


gabrielesvelto (approved):
LGTM

PR #199, opened 2026-08-10
Isolated Processes 11: Use a user-chosen handler for drop failures

marti4d:
Logging drop failures is generally a good idea... Except that logging in the executor could be a disaster if a logger has already been installed that tries to print to STDERR.

This allows the user to install their own optional Drop failure handler. In an executor process, this simply wouldn't ever be used, so the executor will always do nothing when a Drop fails. Unfortunate, but not much else we can do.

### DIFF
--- a/crates/linux/process-backend/src/drop_fail_handler.rs
+++ b/crates/linux/process-backend/src/drop_fail_handler.rs
@@ -1,0 +1,31 @@
+use core::sync::atomic::{AtomicPtr, Ordering};
+
+static HANDLER: AtomicPtr<Handler> = AtomicPtr::new(core::ptr::null_mut());
+
+pub struct Handler {
+    pub(crate) f: for<'a> fn(core::fmt::Arguments<'a>),
+}
+
+impl Handler {
+    pub const fn new(f: for<'a> fn(core::fmt::Arguments<'a>)) -> Self {
+        Self { f }
+    }
+    pub fn install_global(&'static self) {
+        HANDLER.store((self as *const Handler).cast_mut(), Ordering::Release);
+    }
+}
+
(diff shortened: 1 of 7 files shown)


crates/linux/process-backend/src/drop_fail_handler.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub struct Handler {
+    pub(crate) f: for<'a> fn(core::fmt::Arguments<'a>),
+}
+
+impl Handler {
+    pub const fn new(f: for<'a> fn(core::fmt::Arguments<'a>)) -> Self {
+        Self { f }
+    }
+    pub fn install_global(&'static self) {
```
gabrielesvelto, line 36: It'd be worth documenting that this is not synchronized, so it's up to the caller to sync appropriately if this is used in a multi-threaded environment.

gabrielesvelto (approved):

crates/linux/process-backend/src/drop_fail_handler.rs
marti4d: I'm curious what you mean by "not synchronized"? It's using atomic acquire-release, so multiple threads should be fine to use it... Unless I'm missing something?

gabrielesvelto: Yeah, I meant that this is a global and if it's set more than once by different pieces of code the last one will (silently) win.

marti4d: Ah, good point -- I hadn't considered that some folks might expect "once init" behavior instead of update behavior. I added some prose to the beginning of the module to explain things better :)

PR #200, opened 2026-08-10
Isolated Processes 12: Refactor Result, visiblity, wrappers

marti4d:
There are about to be a lot of methods returning Result<T, Error>, so let's just start using some specialized results to avoid repeating ourselves.

Also, some of the visiblity on things could be cleaned up.

And let's centralize some of the wrappers for things like FDs so the remote end can reuse them.

There's not much to this, but it makes other reviews noisy so I thought it best to separate it out.

### DIFF
--- a/src/linux/auxv/mod.rs
+++ b/src/linux/auxv/mod.rs
@@ -97,7 +97,7 @@ impl AuxvDumpInfo {
 
         let auxv_path = format!("/proc/{pid}/auxv");
         let auxv_file = process_inspector
-            .read_file(&auxv_path)
+            .read_file(auxv_path.clone().into())
             .map_err(|e| AuxvError::OpenError(auxv_path, e))?;
 
         for pair_result in ProcfsAuxvIter::new(BufReader::new(auxv_file)) {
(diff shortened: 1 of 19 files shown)


src/linux/auxv/mod.rs
```diff
@@ -97,7 +97,7 @@ impl AuxvDumpInfo {
 
         let auxv_path = format!("/proc/{pid}/auxv");
         let auxv_file = process_inspector
-            .read_file(&auxv_path)
```
marti4d, line 100: So, if you're confused about why I had to do this (explicitly make every API require a `PathBuf` instead of taking a `impl Into<PathBuf>`), it's because we're about to make `ProcessInspector` a dyn Trait -- And dyn Traits can't have generic arguments.

Unfortunately, this means I have to go to every call site and write some variation of `.into()`, `PathBuf::from()`, `path.clone()`, etc.

Shitty, but... Whatcha gonna do?

marti4d:
@gabrielesvelto - This is a lot of small changes, but honestly there's not a whole lot to see here - Mostly just replacing a bunch of uses of `Result<T, Error>` with a specialized typedef, tightening up visibility of symbols, fixing the ptrace unit tests to run one-at-a-time (as it turns out, they all race each other to create and destroy threads, but we often don't see it because the threads roughly all start at the same time and run fast enough to not witness each other's destruction), and I finally make the soft error test less sensitive (it was annoying that it blew up any time anyone changed the structure of error enums).

gabrielesvelto (approved):
LGTM

PR #202, opened 2026-08-10
Isolated Processes 14: Add traits for process-backend implementations

marti4d:
We are about to have a second set of implementations for the various process-backend types (the remote kind).

We can, of course, implement the `ProcessInspector` traits for each of these, individually, but I think it makes good sense from an API perspective for process-backend to establish "this is my contract" rather than having two separate contracts for `local` and `remote` implementations.

Of course, the nice part is that now `ProcessInspector` can be blanket implementated for all `Backend` types at once.


gabrielesvelto (approved):
LGTM

PR #203, opened 2026-08-10
Isolated Processes 15: Modify API to use serializable types

marti4d:


marti4d:
If you are curious about the motivation here, read the top of the `src/regs.rs` file, but basically: `libc` is rather inconsistent about defining a lot of the constants/structs for the different register sets, it sometimes doesn't define them even though the kernel technically supports them, and it varies depending on whether you're using glibc, bionic, or musl.

They also don't implement `Default`, `Deserialize` or `Serialize`, even though they all could because zero would be a reasonable default for all fields, and they could all easy implement the `serde` traits.

So... I basically went through the kernel, duplicated them myself into the regs.rs file, and cited the actual kernel source code for each structure.

The nice thing, now, is that I have a new `unsafe trait PtraceRegisterSet { ... }`, that must ensure that the ptrace note, request, kernel size, and output type all match each other. That allows the API calls themselves to be made safe for any of the types that I've manually checked against the kernel code.

crates/linux/process-backend/src/regs.rs
```diff
@@ -1,138 +1,448 @@
...
+/// 3. `KERNEL_SIZE <= size_of::<Output>()`.
+///
+/// 4. Every byte pattern the kernel may leave in the first `KERNEL_SIZE`
+///    bytes, combined with the untouched bytes of `Output::default()`,
+///    must constitute a valid `Output`, including when the syscall fails
+///    after a partial write.
+///
+/// 5. `Output` must not require drop glue.
+pub(crate) unsafe trait PtraceRegisterSet {
+    const NOTE: usize;
```
schopin-mozilla: suggestion: use a newtype for `NOTE` values (or just a type alias)

Until I understood that `NOTE` was just an arbitrary kernel-defined ID contrary to `KERNEL_SIZE`, and that `NT_PRSTATUS` was such a note, the first guarantee was a bit hard to parse.

crates/linux/process-backend/src/regs.rs
```diff
@@ -1,138 +1,448 @@
...
+    const LEGACY_REQUEST: Option<PtraceRequestType>;
```
schopin-mozilla, line 59:
```suggestion
    /// Legacy ptrace request to fall back on if PTRACE_GETREGSET fails
    const LEGACY_REQUEST: Option<PtraceRequestType>;
```

crates/linux/process-backend/src/regs.rs
```diff
@@ -1,138 +1,448 @@
...
         pub fpscr: u32,
+
+        // Note: not explicit in the kernel's declaration of this struct.
+        //
+        // Technically, the kernel copies this struct as-if it's 65 u32s even though its
+        // C layout puts it at 66 u32s with padding. But, it's permitted to pass in a buffer
+        // larger than necessary to ptrace(PTRACE_GETREGSET), so that's why code that uses the
+        // C definition of the structure works just fine.
+        #[serde(skip)]
+        pub __padding: u32,
```
schopin-mozilla: question(not-blocking): why do we need this padding?

It doesn't hurt so it's fine, but AFAICT we're bypassing the libc to use straight syscalls anyway, so why are we reproducing their ABI?

crates/linux/process-backend/src/local/mod.rs
```diff
@@ -204,87 +198,59 @@ impl Backend {
...
-        {
-            Err(Error::NotSupported)
-        }
-        #[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
-        {
-            const PTRACE_GETFPREGS: PtraceRequestType = 14;
-            unsafe { self.ptrace_getregs::<FpRegs>(PTRACE_GETFPREGS, _tid) }
-        }
-    }
+    fn ptrace_getregs<T: PtraceRegisterSet>(&self, tid: libc::pid_t) -> Result<T::Output> {
```
schopin-mozilla:
```suggestion
    fn ptrace_getregs_legacy<T: PtraceRegisterSet>(&self, tid: libc::pid_t) -> Result<T::Output> {
```
The `ptrace_getregs()` name implies we're using the `PTRACE_GETREGS` request, which is not necessarily the case.

schopin-mozilla (approved):
Altogether LGTM, just a few suggestions to improve readability.

crates/linux/process-backend/src/regs.rs
marti4d: Good suggestion! Done!

crates/linux/process-backend/src/regs.rs
marti4d: Good one! Done.

crates/linux/process-backend/src/local/mod.rs
marti4d: Yep -- That's misleading. Done!

crates/linux/process-backend/src/regs.rs
marti4d: Honestly: Consistency. Every other structure here has all of its padding explicitly declared, so I figured I would follow suit with this.

But it's true that there's another way to look at it, which is that all these other structs guarantee that the kernel will actually zero out their padding bytes, and this struct doesn't - It's truly uninitialized padding. Of course -- We zero it out when we default it, but yeah... I guess the risk with declaring it like this is that this could lead to UB:

```
let mut vfp = MaybeUninit::<user_vfp>::uninit();
let rv = unsafe { ptrace(..., &mut vfp) };
// ...
unsafe { vfp.assume_init() } // UB - last padding field isn't initialized.
```

So, now that I think about it -- I've talked myself out of doing it this way 😅

marti4d: Alrighty -- I removed the padding and integrated the comment into the comment for the overall struct.

gabrielesvelto (approved):
LGTM, this is such a big improvement in terms of readability and robustness over what we had and a huge leap over the breakpad code we're coming from.

marti4d:
> LGTM, this is such a big improvement in terms of readability and robustness over what we had and a huge leap over the breakpad code we're coming from.

Thanks @gabrielesvelto - That's kind of you to say 😊

PR #204, opened 2026-08-10
Isolated Processes 16: Implement Remote Backend

marti4d:


marti4d:
@gabrielesvelto So... Even though I tried to break all the previous changes up, this change is... A lot still.

If you want, I'd be happy to do a Zoom call or something and give you a tour/overview of the code from a high-level so you can at least get an idea of what everything is, why it's there, etc.

marti4d:
Closing because I am splitting this up further

PR #206, opened 2026-08-17
Android CI improvements

schopin-mozilla:
This PR aims at two things:
* Run Clippy on Android code (similar idea to #197), and fix what fell out
* Make the Android timeouts a bit more generous as that platform test runner isn't as spry as others.

It's all split off my and Chris' work in #195


marti4d (approved):
LGTM!

PR #208, opened 2026-08-17
tests: get a test on module list properties

schopin-mozilla:
We're about to do some extensive work on the module list, and the existing integration tests only cover the "which modules do we have" part.

Let's add a new test that encodes some other properties of the list: it starts with the executable, and the modules don't overlap. This should make working on that specific feature feel a bit more confident.

Split off the #195 work.


marti4d (approved):
Awesome!

PR #209, opened 2026-08-17
Fix ptrace race on Android tests

schopin-mozilla:
The main goal of this PR is to avoid a race when running multiple tests in parallel where they all try to ptrace() the parent and block each other out. I threw in a few output enhancements that I had to do in order to diagnose the issue.

### DIFF
--- a/.cargo/android-runner.sh
+++ b/.cargo/android-runner.sh
@@ -10,4 +10,4 @@ TARGET=$(printf '%s' "$BINARY" | sed -E 's#(.*/)?target/([^/]+)/.*#\2#')
 # Make sure to run the following to copy the test helper binary over.
 # cargo run --target "$TARGET" --bin test
 adb push "$BINARY" "/data/local/$BINARY"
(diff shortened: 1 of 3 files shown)


marti4d (changes requested):
Oh no... We had a mid-air collision on this too! And again -- I just merged this in before I saw yours!

https://github.com/rust-minidump/minidump-writer/blob/main/tests/ptrace_dumper.rs#L47

Sorry about both of these - My fault. I haven't really been communicating that I had to fix these things as part of making my process inspection changes work.

It's funny -- We both ran into some of the same issues with testing doing our work. Our changes aren't even that dissimilar lol.

It looks like you might have taken a less shotgun-y approach here though - Yours might be clearly better. What do you think? Fix the merge conflicts and still do this instead?

schopin-mozilla:
Sure! I do love that not only did we end up with the same solution, but we *also* had the same side-quest of getting the output sorted out!
Rebased, and I took the liberty of removing your approach to avoid duplication.

marti4d (approved):
Perfect 💯

Thanks for doing this -- My way was... significantly crappier than this lol.

PR #210, opened 2026-08-17
linux: change some APIs to use mem address rather than mapping index

schopin-mozilla:
Those APIs are used to get data that's typically module-related, so having them take in an index into the *mappings* list can get pretty awkward once we start separating those concerns.

As a bonus, I find the client code to be more readable, although YMMV.

Split off #195 (I'm running out of small pieces I can spin out independently, sadly)


marti4d (approved):
Lol thanks for doing this -- Yep, I agree. The old API was awkward, and this is much better.

PR #212, opened 2026-08-18
linux: separate mappings list from modules list

schopin-mozilla:
While both things are related to each other, and for now we're using the
former to generate the latter, they are essentially two different sets
of data, meant to answer two different questions. Since the endgoal of
this patch series is to use a *different* data source for the module
info, the first step is to actually make it its own entity.

My design goal for this is to only construct ModuleInfo objects once we
know they're correct (or as correct as they are), which is why we moved
the user-supplied mapping processing right within the process, and it
all goes through an intermediate struct. There's a couple of advantages
to this: we have all module-related logic stored in a single place,
including arbitration between user-provided and process-derived data,
and if you get a ModuleInfo object you know it's the same data that will
get written out.

An unplanned benefit of this is that mappings don't need to know about
bias anymore, which is nice since it's an ELF notion, not a generic
mapping one.

The final result should be the same as before, except for one case:
conflicts between process-derived and user-provided data is now
explicitly handled, with the latter given priority.

Additionally, errors during SONAME extractions are now reported in the
soft error stream rather than completely ignored. It's a bit tangential,
but was entangled too much with the rest to be split out into its own
commit.

(description lifted from the first and most meaty commit, the second one being just some cleanup.)
(split off #195)

### DIFF
--- a/src/linux/minidump_writer/mappings.rs
+++ b/src/linux/minidump_writer/mappings.rs
@@ -1,107 +1,121 @@
-use super::{super::maps_reader::MappingInfo, *};
+use {
+    super::*,
+    crate::module_list::{ModuleListError, ModuleSource},
+    std::ffi::OsString,
+};
 
 #[derive(Debug, Error, serde::Serialize)]
 pub enum SectionMappingsError {
     #[error("Failed to write to memory")]
     MemoryWriterError(#[from] MemoryWriterError),
-    #[error("Failed to get effective path of mapping ({0:?})")]
-    GetEffectivePathError(MappingInfo, #[source] MapsReaderError),
+    #[error("Errors occurred while inspecting module {}", .name.to_string_lossy())]
+    ModuleErrors {
+        name: OsString,
+        #[source]
+        errors: ErrorList<ModuleListError>,
+    },
 }
 
 impl MinidumpWriter {
-    /// Write information about the mappings in effect. Because we are using the
-    /// minidump format, the information about the mappings is pretty limited.
-    /// Because of this, we also include the full, unparsed, /proc/$x/maps file in
-    /// another stream in the file.
+    /// Write information about the modules loaded into the process. Because we
+    /// are using the minidump format, the information about the modules is
+    /// pretty limited. Because of this, we also include the full, unparsed,
+    /// /proc/$x/maps file in another stream in the file.
     pub fn write_mappings(
-        &mut self,
+        &self,
         buffer: &mut DumpBuf,
+        mut soft_errors: impl WriteErrorList<SectionMappingsError>,
     ) -> Result<MDRawDirectory, SectionMappingsError> {
-        let mut modules = Vec::new();
+        let mut raw_modules = Vec::new();
 
-        // First write all the mappings from the dumper
-        for mapping in &self.mappings {
-            // If the mapping is uninteresting, or if
-            // there is caller-provided information about this mapping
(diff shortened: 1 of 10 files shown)


schopin-mozilla:
Ping @marti4d for review. Sorry, it's a bit massive :/

src/linux/minidump_writer/mappings.rs
```diff
@@ -1,107 +1,121 @@
...
+        if !raw_modules.is_empty() {
+            let module_list =
+                MemoryArrayWriter::<MDRawModule>::alloc_from_iter(buffer, raw_modules)?;
+            dirent.location.data_size += module_list.location().data_size;
         }
 
         Ok(dirent)
     }
+
+    /// What it says on the tin. Returns None if the module is deemed "uninteresting".
```
gabrielesvelto: nit: Describe what uninteresting modules are. I know but we should make sure other people are able to easily figure out what's going on.

src/linux/module_list.rs
```diff
@@ -0,0 +1,471 @@
...
+    // one, then we need to make it be first. This is because the minidump
+    // format assumes the first module is the one that corresponds to the main
+    // executable (as codified in processor/minidump.cc:MinidumpModuleList::GetMainModule()).
+    if let Some(entry_point) = entry_point
+        && let Some(index) = process_candidates
+            .iter()
+            .position(|c| c.contains_address(entry_point))
+    {
+        process_candidates.swap(0, index);
+    }
```
gabrielesvelto: Just one nit about organizing this type of code: if a long comment is required before a block it's usually a good idea to split it out into a separate function with a descriptive name (e.g. `move_entry_point_module()` or something along the lines in this example). It's OK to move the comment there for further clarification. Generally speaking it makes reading the code easier.

src/linux/module_list.rs
```diff
@@ -0,0 +1,471 @@
...
+            }
+            match self.so_name(process_inspector) {
+                Ok(soname) => soname,
+                Err(e) => {
+                    // Log error and move on to fallback path.
+                    soft_errors.push(e);
+                    None
+                }
+            }
+        });
```
gabrielesvelto: Split this out in a separate function.

src/linux/module_list.rs
```diff
@@ -0,0 +1,471 @@
...
+            // the executable was loaded directly from inside an archive file (e.g., an
+            // apk on Android).
+            // In this case, we append the file_name to the mapped archive path:
+            //   file_name := libname.so
+            //   file_path := /path/to/ARCHIVE.APK/libname.so
+            file_path.push(&file_name);
+        } else {
+            // Otherwise, replace the basename with the SONAME.
+            file_path.set_file_name(&file_name);
+        }
```
gabrielesvelto: Likewise

gabrielesvelto (approved):
Sorry, I meant to have approved this with nits so you can merge when they're addressed

src/linux/module_list.rs
schopin-mozilla: A lot of these were just copied over from the original mapping.rs code, but I suppose I might as well split them out while I'm there! All done.

gabrielesvelto: Yes, and those were in turn lifted from Breakpad's code so I'm eager to see them out :-P

gabrielesvelto: Thanks for addressing this while you were at it

PR #213, opened 2026-08-20
Fix clippy lint broken with new Rust release

marti4d:


marti4d:
@gabrielesvelto This doesn't actually need a code review (trivial change), but it's worth thinking about if we want to always continue the current behavior of having our CI  automatically jump up to the latest version of Rust.

The combination of "CI automatically updates Rust" + "We run clippy and fail on any warning" means that code that was passing tests one day can fail them the next day without anybody explicitly doing anything. Not great...

gabrielesvelto:
> The combination of "CI automatically updates Rust" + "We run clippy and fail on any warning" means that code that was passing tests one day can fail them the next day without anybody explicitly doing anything. Not great...

Good point. I'll open an issue, updating the harness to pick a new version of Rust only after we've verified that it works is a better path forward.

PR #215, opened 2026-08-21
linux: modules: clamp when they overlap other modules

schopin-mozilla:
This isn't a problem yet (barring user-provided data, of course) since
the procmaps approach cannot, by obstruction, generate overlapping
modules, but it will become a problem when we'll be using the DT_DEBUG
data.

Indeed, some objects are disjointed in memory, and because of that you
can have other modules inserted into the gap. In particular, on Linux
arm64, the vDSO routinely lands in the middle of ld.so segments.

It's usually benign since the latter segments are only data, not code,
and we record an error if not.

### DIFF
--- a/src/linux/module_list.rs
+++ b/src/linux/module_list.rs
@@ -29,6 +29,17 @@ pub enum ModuleResolveError {
         name.to_string_lossy()
     )]
     DuplicateUserModule { name: OsString, base_address: usize },
+    #[error(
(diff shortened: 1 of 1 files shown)


schopin-mozilla:
Hmm, I tried to do the Stack PR thing but it failed, apparently it needs to be enabled at the repo level. Oh well.

In this case, DO NOT MERGE before #212 !

gabrielesvelto (approved):
LGTM for the last patch (given the rest will land in the previous PR)

PR #217, opened 2026-08-24
Isolated Processes 17: Implement executor to run syscalls in remote process

marti4d:
@gabrielesvelto I broke the original Isolated Processes 16 into 16,17,18, 19 to make it easier to review in chunks.

Each chunk compiles, but a lot of it is dead code until it's finally assembled in the last chunk.

### DIFF
--- a/crates/linux/process-backend/src/remote/executor/handler.rs
+++ b/crates/linux/process-backend/src/remote/executor/handler.rs
@@ -12,6 +12,7 @@ impl HandlesRequests for &mut Executor<'_> {
     {
         Ok(self.local_backend.pid())
     }
+
     fn ProcessReaderReadAt<'args, 'output>(
         self,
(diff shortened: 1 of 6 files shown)


crates/linux/process-backend/src/remote/executor/handler.rs
```diff
@@ -0,0 +1,246 @@
...
+        self.local_backend.fail_one_syscall_with(errno);
+        Ok(())
+    }
+    fn Quit<'args, 'output>(self, _args: ()) -> Result<()>
+    where
+        Self: 'output,
+    {
+        self.quit = true;
+        Ok(())
+    }
```
gabrielesvelto, line 263: nit: Add a blank line between each function

gabrielesvelto (approved):
LGTM, just one formatting nit

PR #220, opened 2026-08-24
linux: add memory-reading helpers for reading POD values from a process

schopin-mozilla:
Pulled off #195

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1374,6 +1374,7 @@ dependencies = [
  "minidump",
  "minidump-common",
  "minidump-unwind",
+ "plain",
  "process-backend",
  "procfs-core 0.18.0",
  "scroll 0.12.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ memmap2 = "0.9"
 byteorder = "1.4"
 error-graph = { version = "0.1.1", features = ["serde"] }
 failspot = "0.2.0"
+plain = { version = "0.2.3", default-features = false }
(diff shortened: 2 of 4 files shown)


schopin-mozilla:
@marti4d FYI

marti4d (changes requested):
Sorry, but could I please ask you to implement `read_exact` in `process_inspection` and not in `process-backend`?

The issue is that any API that's added to `process-backend` is an API that will need to be remoted to a crashed content process running in a signal handler context -- Which means it can't allocate and has to use fixed-sized buffers.

So all the APIs that are in `process-backend` have been designed to either have output types that are a fairly-small, fixed size (maybe a few hundred bytes), or-else they return a buffer that is dynamically sized but can be truncated from what the user requested (the way that `read()`-like APIs behave, where you can give the user less than they asked for if you don't have the buffer space for it).

So, unfortunately, putting `read_exact()` into `process-backend` breaks that contract - It could make an atomic request that is larger than the remote backend can buffer, causing a guaranteed failure to ever serve the request.

schopin-mozilla:
My initial thought was to just move the current implementation (process_reader::ProcessReader::read_exact_at) into a default method implementation on the reader backend trait, but then I would have had to wire up the errors, etc... which makes any other implementor's life harder, so instead I just duplicated the algorithm where you wanted instead. After all, it was already present in two places!

That's a lie. My initial thought was "there's too many things called the same thing in this mess".

marti4d (approved):
LGTM... Thanks!

PR #221, opened 2026-08-24
linux: document and simplify the dso_debug code

schopin-mozilla:
We use the new `read_pod` to simplify the code.

Split off #195 (sort of)

Do not merge before #220 !!


marti4d (approved):
Nice! And glad to see you fixed up and added all the comments too :)

marti4d:
(As you said -- we should hold off on landing this until #220 is landed)

PR #224, opened 2026-08-25
Make process-backend and process-reader private crates

marti4d:
These two crates were accidentally released on crates.io because we forgot to add the "publish = false" option to Cargo.toml. They are internal crates to minidump-writer and aren't really meant for public consumption, so it's unfortunate to have made that mistake.

This prevents us from accidentally doing this in the future.

(Note that, eventually, we may actually release `process-reader` as an independently-useful crate for the community... But it needs to be buffed up a bit before it's ready for that).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1661,7 +1661,7 @@ dependencies = [
 
 [[package]]
 name = "process-backend"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "libc",
  "process-reader",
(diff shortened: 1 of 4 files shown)


marti4d:
Nevermind -- I see that we can't even publish the main crate on crates.io without these.

I think I'm going to have to rename them to `minidump-writer-process-backend` and `minidump-writer-process-reader` to avoid clogging up `crates.io` with private implementation details of our crate.

PR #227, opened 2026-08-26
linux: dso_debug: add a MainExecutable struct

schopin-mozilla:
This allows us to split out the code in smaller bits, which will come in handy later on.

There's a drive-by fix, where we stop relying on page boundaries and instead compute the actual load bias of the executable. This fixes support for non-4kB page sizes.

### DIFF
--- a/src/linux/dso_debug.rs
+++ b/src/linux/dso_debug.rs
@@ -5,8 +5,11 @@
 
 use {
     super::{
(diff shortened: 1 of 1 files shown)


schopin-mozilla:
Cc @marti4d

schopin-mozilla:
Closing this as I'll have to push this to a branch on this repo rather than my fork for stacked PRs.

PR #229, opened 2026-09-01
linux: dso_debug: add a MainExecutable struct

schopin-mozilla:
This allows us to split out the code in smaller bits, which will come in
handy later on.

There's a drive-by fix, where we stop relying on page boundaries and
instead compute the actual load bias of the executable. This fixes
support for non-4kB page sizes.

---


src/linux/dso_debug.rs
```diff
@@ -58,6 +60,138 @@ pub enum SectionDsoDebugError {
...
+    fn compute_bias(
+        program_headers: &[ProgramHeader],
+        program_header_table_address: usize,
+    ) -> RendezvousResult<usize> {
+        let table = program_headers
+            .iter()
+            .find(|hdr| hdr.p_type == PT_PHDR)
+            .ok_or(RendezvousError::ProgramHeaderTableNoSelf)?;
+
+        Ok(program_header_table_address - usize::try_from(table.p_vaddr).unwrap())
```
gabrielesvelto: nit: For this and the other instance where we turn a virtual address into an `usize` I'd use something like `.expect("Pointer size mismatch between dumper and target")` rather than a bare `.unwrap()`. Failure here would be possible only if one tried to use a 32-bit executable to dump a 64-bit process which is a really odd situation.

src/linux/dso_debug.rs
```diff
@@ -58,6 +60,138 @@ pub enum SectionDsoDebugError {
...
+
+        let entries: Vec<Dyn> = memory_reader
+            .read_pod_vec(address, size / std::mem::size_of::<Dyn>())
+            .map_err(RendezvousError::ReadDynamicSectionFailed)?;
+
+        if entries.last() != Some(&Dyn::default()) {
+            return Err(RendezvousError::DynamicSectionMissingTerminator);
+        }
+
+        Ok((address, entries))
```
gabrielesvelto, line 169: Nice error handling here, I'm sure we'll find weird stuff in the wild.

gabrielesvelto (approved):
LGTM, just one small nit

src/linux/dso_debug.rs
schopin-mozilla: So, this ended up being a bit more involved than originally thought due to dso_debug and module_reader using slightly different types for their ELF data, which is why this was actually fixed in a new patch (#241)

PR #230, opened 2026-09-01
linux: dso_debug: validate RDebug and LinkMap data when iterating them

schopin-mozilla:
This data comes from a process that has potentially corrupted memory,
it's probably a good idea to validate the data as much as possible.

We also abstract LinkMap over the Iterator trait to make it easier to
reuse it later on, e.g. for module list generation.

---


src/linux/dso_debug.rs
```diff
@@ -262,6 +270,78 @@ pub struct RDebug {
...
+        if self.address == 0 {
+            return None;
+        }
+
+        let link = self
+            .reader
+            .read_pod::<LinkMap>(self.address)
+            .map_err(RendezvousError::ReadLinkMapEntryFailed);
+        let Ok(link) = link else {
+            self.address = 0;
```
gabrielesvelto: Just a stylistic nit, while this is just one line it's better to put it in a separate function with a descriptive name like `terminate_iterator()` or something to make the intent clear.

src/linux/dso_debug.rs
```diff
@@ -262,6 +270,78 @@ pub struct RDebug {
...
+            return Some(link);
+        };
+
+        if self.first_node {
+            // We were handed `r_map`, so nothing should precede us.
+            if link.l_prev != 0 {
+                self.address = 0;
```
gabrielesvelto: Likewise

gabrielesvelto (approved):
LGTM, just a small nit

PR #231, opened 2026-09-01
linux: dso_debug: move LinkMap name extraction into its own function

schopin-mozilla:
We'll need this elsewhere as well. We also use the read_until helper
rather than rolling our own, thus avoid the arbitrary 256 char limit.

---


gabrielesvelto (approved):
LGTM

PR #232, opened 2026-09-01
linux: Enumerate mappings *after* suspending threads

schopin-mozilla:
At this stage, the main effect of suspend_threads is to attach ptrace,
which we'll need when we start reading the target's memory to build the
module list.

Co-Authored-By: Simon Chopin <schopin@mozilla.com>

---


gabrielesvelto (approved):
LGTM

PR #233, opened 2026-09-01
linux: Add from_debugger_rendezvous to get module candidates from DT_DEBUG

schopin-mozilla:
We use the dso_debug data source, enhanced with data mined out of
module_reader to create an alternative module list source. Since the
output data format is a bit limited, each candidate spans all of the
module's segments.

The main executable needs a bit of massaging as it's sometimes unnamed
and we can't really have that, hence the pid-based fallback.

Not wired up yet.

Co-authored-by: Chris Martin <marti4d@live.ca>

---


src/linux/module_list.rs
```diff
@@ -534,3 +574,244 @@ mod tests {
...
     }
 }
+
+/// Derives the module list by walking the dynamic linker's debugger rendez-vous.
+///
+/// The debugger rendez-vous is reachable through a field in the main program's
+/// headers, hence the need for `process_inspector`, `program_header_table_address`
+/// and `program_header_count`, while `pid` is there in order to get a name for the
+/// main program's module, which isn't necessarily given in the DT_DEBUG data (glibc).
+pub fn from_debugger_rendezvous(
```
gabrielesvelto, line 584: This only needs to be visible within the crate, I don't think we need it as a public interface

src/linux/module_list.rs
```diff
@@ -534,3 +574,244 @@ mod tests {
...
+        &memory_reader,
+        program_header_table_address,
+        program_header_count,
+    )?;
+
+    // Throughout the code stack we're assuming everything is PIE. Notably,
+    // module_reader would require substantial changes to accomodate for
+    // non-PIE executables reliably. Let's just error out rather than return
+    // garbage.
+    if main_executable.load_bias == 0 {
```
gabrielesvelto, line 603: nit: Add an `is_pie()` function to `MainExecutable`

src/linux/module_list.rs
```diff
@@ -534,3 +574,244 @@ mod tests {
...
+        .program_headers()
+        .map_err(ModuleCandidateError::ReadModuleFailed)?;
+
+    // `p_vaddr` is relative to the module's load bias.
+    let mut segments: Vec<Segment> = program_headers
+        .iter()
+        .filter(|hdr| hdr.p_type == PT_LOAD)
+        .map(|hdr| {
+            let start = load_bias + usize::try_from(hdr.p_vaddr).unwrap();
+            let end = start + usize::try_from(hdr.p_memsz).unwrap();
```
gabrielesvelto, line 705: Like in the other patch I'd like something more explicit for ELF address -> host addresses conversions. Maybe even a helper like `elf_addr_to_usize()` which `expect()`s with a descriptive message.

src/linux/module_list.rs
gabrielesvelto, line 584: This can be crate-visible, I don't think we need to expose it publicly, right?

src/linux/module_list.rs
gabrielesvelto, line 603: Add an `is_pie()` function to `MainExecutable`

src/linux/module_list.rs
gabrielesvelto, line 705: Like in the other patch in the stack I'd like this to be more explicit given it's something that could fail only under bizarre conditions. Maybe something like a helper `elf_addr_to_usize()` that `expect()`s with an error message such as "target and source processes differ in pointer-size" or something like that. In case someone is mad enough to try and dump a 64-bit process from a 32-bit one.

src/linux/module_list.rs
```diff
@@ -534,3 +574,244 @@ mod tests {
...
+}
+
+fn page_size() -> usize {
+    let page_size = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+    assert!(
+        page_size > 0,
+        "somehow we weren't able to get the page size"
+    );
+    usize::try_from(page_size).unwrap()
+}
```
gabrielesvelto, line 804: This is something we've already read at the top-level, and having to pass the pid down to read the executable name got me thinking... maybe it would be better to have a struct `DebuggerRendezVousReader` struct with the global info and attach the functions to it, WDYT?

src/linux/module_list.rs
```diff
@@ -534,3 +574,244 @@ mod tests {
...
+
+fn align_up(val: usize, align: usize) -> usize {
+    let result = val / align * align;
+    if !val.is_multiple_of(align) {
+        result + align
+    } else {
+        result
+    }
+}
```
gabrielesvelto, line 817: [`next_multiple_of()`](https://doc.rust-lang.org/std/primitive.u64.html#method.next_multiple_of) should do the trick

gabrielesvelto (changes requested):

src/linux/module_list.rs
schopin-mozilla: IIRC I did that to avoid unused warnings since it's not wired yet. I replaced it with a temporary inhibitor that is then removed in the next PR.

src/linux/module_list.rs
schopin-mozilla: I went the helper route, and used it in subsequent patches as well.

gabrielesvelto (approved):
LGTM, ty!

PR #234, opened 2026-09-01
linux: optionally get module data from the debugger rendez-vous

### LINKED ISSUE #142: Use the debugger rendez-vous to populate the module list on Linux

schopin-mozilla:
Wire up the debugger rendezvous as an alternative data source for module candidates, complete with config knob, fallback paths, and tests.

Fixes #142 

---


gabrielesvelto (approved):
LGTM

PR #235, opened 2026-09-01
linux: don't walk the rendez-vous while the linker is mid-update

schopin-mozilla:
While playing around with edge cases scenarii, one of them stood out:
what would happen if there was a crash *during* a dlopen() call? That's
what this patch is about.

Luckily for us, the libc folks know their stuff, and have a marker field
in the RT_DEBUG structure: `r_state`. It is present in both bionic and
glibc, with the same standard values, so we can probably assume it's
fairly standard, probably part of the SVR4 ABI – not that it's
documented anywhere besides random blog articles that lamented at the
same thing.

In any case, we can now bail early if the linker doesn't explicitly mark
its data as consistent. Note that this impacts the DSO data dump, that
will now be empty rather than filled with garbage, as well as the module
stream that'll fall back to procmaps.

---


schopin-mozilla:
Note that this PR comes in this late in the stack because the tests for it are using things from previous patches. The fix itself could have been moved way earlier.

gabrielesvelto (approved):
LGTM

Issue #222, opened 2026-08-25
process-backend: Don't swallow the original error in `local::Backend::get_gen_regs/get_fp_regs`


marti4d:
We can see from Firefox [Bug 1245570](https://bugzilla.mozilla.org/show_bug.cgi?id=1245570) the consequences of doing this. We can see that `getfpregset()` failed on an `arm-linux-androideabi` device, which swallowed the error and called `getfpregs()`, overwriting the original error with the practically-useless `Error::NotSupported` -- So now we have no idea why `ptrace(PTRACE_GETREGSET, NT_ARM_VFP)` failed.

(Even though the honest answer is that the kernel is so old that it probably didn't exist yet).

Issue #223, opened 2026-08-25
Consider treating a lot of hard errors around information gathering as soft errors


marti4d:
[Firefox Bug 1245570](https://bugzilla.mozilla.org/show_bug.cgi?id=1245570) shows that we might be, perhaps, a bit over-reacting to how we handle a failure to obtain the floating point register values.

Currently, if we fail to obtain both the general-purpose registers and the floating-point registers as part of writing the `ThreadListStream` entry, we bomb out with a hard error [here](https://github.com/rust-minidump/minidump-writer/blob/45f5459790d264098346343f0ad58d008c386fa5/src/linux/minidump_writer/thread_list_stream.rs#L161), which then also causes us to give up on writing the minidump altogether [here](https://github.com/rust-minidump/minidump-writer/blob/45f5459790d264098346343f0ad58d008c386fa5/src/linux/minidump_writer/mod.rs#L348).

This issue is to investigate and discuss whether we can have **a bit more chill** when it comes to error handling for either or both of these steps. Perhaps we can still push forward writing the `ThreadListStream` even if we can't read some of the registers... Or maybe we can't, but maybe we should not treat a failure to write the `ThreadListStream` as a total loss for writing the minidump.

More generally, we currently bomb out if we fail to write a lot of different streams: `ModuleListStream`, `MemoryListStream`, `ExceptionStream`, etc -- The list goes on. It might be worth taking inventory of all the stuff written in [`MinidumpWriter::write_dump()`](https://github.com/rust-minidump/minidump-writer/blob/45f5459790d264098346343f0ad58d008c386fa5/src/linux/minidump_writer/mod.rs#L315) and differentiating "if we can't write this stream, the whole minidump is worthless and we might as well not send it" versus "we could still do useful things with a minidump that lacks this stream".

Curious to hear some of your thoughts, @gabrielesvelto .

marti4d:
@schopin-mozilla made a good point -- At the very least, we should always send the Soft Errors stream, and the soft error stream should ideally even contain any hard error that stopped us from gathering information in the rest of the minidump.

So even if we hit a hard error that makes the minidump completely useless otherwise, we still have the soft errors and hard error that caused us to stop trying to make a useful minidump.

PR #238, opened 2026-09-02
Isolated Processes 15.5: Treat non-essential getreg errors as soft

marti4d:


marti4d:
Related to Issue #223

Fixes Issue #222

gabrielesvelto (approved):
LGTM, I wonder what we're going to discover with this. Every time we add some more granular error-reporting surprises come out of it.

PR #241, opened 2026-09-14
linux: dso_debug: use width-agnostic ELF types in its APIs

schopin-mozilla:
The goblin crate exposes a set of types that can represent headers of
both 32-bit and 64-bit ELF objects, and that's what module_reader uses.
In order to facilitate interaction between those two Rust module, let's
use that rather than the hardcoded native variants in our pub(crate)
APIs. The module still uses the native versions at parsing time, so it's
not portable, but it will make the module list work easier.

---


gabrielesvelto (approved):
Nice cleanup

